PR #4, opened 2022-07-04
[CMS-754] Use new pantheon-systems/composer-scaffold package.

kporras07:
Depends on https://github.com/pantheon-systems/wordpress-bedrock-integrations/pull/3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,8 +42,8 @@
     "roots/wp-password-bcrypt": "1.1.0",
     "wpackagist-theme/twentytwentytwo": "^1.2",
     "pantheon-systems/wordpress-bedrock-integrations": "^2.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -67,10 +67,7 @@
       "web/app/themes/{$name}/": ["type:wordpress-theme"]
     },
     "wordpress-install-dir": "web/wp",
-    "drupal-scaffold": {
-      "allowed-packages": [
-        "pantheon-systems/wordpress-bedrock-integrations"
```
jspellman814, line 72: @kporras07 Do we still need to define which packages are allowed to use the scaffold?

jspellman814: Thank you for clarifying! That makes sense to me.

jspellman814 (approved):

PR #8, opened 2022-07-15
[CMS-801] Adjust wp-config

jspellman814:
Fixes WP-CLI error `Fatal error: Cannot redeclare wp_get_server_protocol() (previously declared in /code/web/wp/wp-includes/load.php:16) in /code/web/wp/wp-includes/load.php on line 15`.

Terminus:

Screen Shot 2022-07-15 at 2 25 02 PM

btool shell:

Screen Shot 2022-07-15 at 2 27 09 PM

### DIFF
--- a/web/wp-config-pantheon.php
+++ b/web/wp-config-pantheon.php
@@ -168,26 +168,3 @@ if ( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) ):
 	endif;
 
 endif;
-
-/*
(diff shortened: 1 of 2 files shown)


CdrMarks (approved):

jazzsequence (approved):
This works for me on my Bedrock multisite install. :shipit: 

I noticed the `config/application.php` is loaded only if a local `wp-config-pantheon.php` does not exist. That's probably most correct now, but I think we can revisit that logic in CMS-802.

PR #9, opened 2022-07-19
[CMS-802] More closely align config management to roots/bedrock

jspellman814:


config/application.php
```diff
@@ -121,10 +110,59 @@
...
+ * For developers: WordPress debugging mode.
+ *
+ * Change this to true to enable the display of notices during development.
+ * It is strongly recommended that plugin and theme developers use WP_DEBUG
+ * in their development environments.
+ *
+ * You may want to examine $_ENV['PANTHEON_ENVIRONMENT'] to set this to be
+ * "true" in dev, but false in test and live.
+ */
+if (!env('WP_DEBUG')) {
```
jazzsequence, line 151: I'm assuming the `env()` function works like this: https://gist.github.com/mattstauffer/59170bc2cb6318ab6d46

In that case, if there was no WP_DEBUG value, it would also return false, making the conditional true, and disabling WP_DEBUG. While this is not necessarily a problem, it would execute line 152 if

* no value was set for WP_DEBUG
* a value was set for WP_DEBUG and it was `false`

Since WP_DEBUG _defaults_ to `false`, perhaps this would be better if the conditional was flipped, e.g.

```php
if ( env('WP_DEBUG') === true ) {
    Config::define( 'WP_DEBUG', true );
}
```

web/wp-config.php
```diff
@@ -4,37 +4,6 @@
...
-if ( ! defined( 'ABSPATH' ) ) {
-	define( 'ABSPATH', dirname( __FILE__ ) . '/' );
-}
-
-/** Sets up WordPress vars and included files. */
-require_once( ABSPATH . 'wp-settings.php' );
\ No newline at end of file
+require_once dirname(__DIR__) . '/vendor/autoload.php';
+require_once dirname(__DIR__) . '/config/application.php';
+require_once ABSPATH . 'wp-settings.php';
```
jazzsequence, line 9: If a user is adding their `wp-config.php` customizations, where should they be added? `config/application.php`? `wp-config.php`? If `wp-config`, we should have a `/* That's all, stop editing! Happy Pressing. */` line so they know where they should put their customizations. I think after the `require_once ABSPATH . `wp-settings.php` may be too late.

.env.pantheon
```diff
@@ -0,0 +1,13 @@
...
+
+# Pantheon sets these values for you. If you want to shuffle them you can do so via your dashboard.
+AUTH_KEY=$_ENV['AUTH_KEY']
+SECURE_AUTH_KEY=$_ENV['SECURE_AUTH_KEY']
+LOGGED_IN_KEY=$_ENV['LOGGED_IN_KEY']
+NONCE_KEY=$_ENV['NONCE_KEY']
+AUTH_SALT=$_ENV['AUTH_SALT']
+SECURE_AUTH_SALT=$_ENV['SECURE_AUTH_SALT']
+LOGGED_IN_SALT=$_ENV['LOGGED_IN_SALT']
+NONCE_SALT=$_ENV['NONCE_SALT']
```
jazzsequence, line 13: Would we want to put WP_HOME and WP_SITEURL in here, too, instead of in `application.php`? 
this mostly applies if we are advising people to edit `application.php` which, as I say elsewhere, I think we should maybe not do and instead instruct them to edit `wp-config.php`. My thought here is that Bedrock's addition of the `/wp` directory breaks a lot of stuff, and if a user saw it and said "that's not right" or "I don't want to log in at /wp/wp-admin, I'll just change this" they could break their site.

config/application.php
```diff
@@ -1,7 +1,6 @@
 <?php
 /**
- * Your base production configuration goes in this file. Environment-specific
- * overrides go in their respective config/environments/{{WP_ENV}}.php file.
+ * Your base production configuration goes in this file.
```
jazzsequence, line 3: Do we want to retain support for environment-level config files? Or would all of that stuff be identified by the PANTHEON_ENVIRONMENT anyway, and therefore be handled in other ways than what Bedrock does?

config/application.php
```diff
@@ -25,35 +24,40 @@
...
     if (!env('DATABASE_URL')) {
         $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
     }
 }
 
-/**
- * Set up our global environment constant and load its config first
- * Default: production
- */
-define('WP_ENV', env('WP_ENV') ?: 'production');
```
jazzsequence, line 49: Is this constant used by anything? Should we define it based on PANTHEON_ENVIRONMENT?

config/application.php
```diff
@@ -96,23 +99,9 @@
 /**
  * Custom Settings
  */
-Config::define('AUTOMATIC_UPDATER_DISABLED', true);
```
jazzsequence, line 99: We still want to disable automatic updates, don't we? Updates would come from the Bedrock upstream.

jazzsequence (approved):
All of this works in testing local and validating on a new single site. There are a host of things that break on multisite (and break in new and exciting ways), but that wouldn't block pushing out a version of this and saying that we support single sites only for now.

There's a number of questions/things to discuss and decide on that I call out. I also am wondering about dev/test/local configurations and how users might specify things in different environments...would they make a `.env.local` and if so, would that override the default environment variables? If not, I think we should keep at least examples of other environments (like the deleted environments files) and support those if they exist and if we're in that environment. I can imagine wanting things like `WP_DEBUG` or `SAVEQUERIES` to be true on dev/test but not on prod.

.env.pantheon
jspellman814: I did [try this](https://github.com/pantheon-systems/wordpress-composer-managed/pull/9/commits/1b7c26fe39d54319d0ac963b42dcbf9a534c1bf2), but was getting a fatal syntax error (don't remember what is was off the top of my head) when trying to load the site. Maybe you'll notice what I did wrong 🤞.

Also team edit `wp-config`, and am working on some comments to indicate that to users.

jspellman814:
@jazzsequence I spoke to Greg about this in my 1:1 and am going to be making some changes based on his feedback, but in general he's on board with our approach here.

jspellman814:
I've marked most of the previous comments as resolved since the latest commits touch a lot of the same things and will need another look. The purpose of these new changes are to reduce the amount of modifications to `config/application.php` to make it easier for us to maintain, and to add a `wp-config-pantheon.php` that is almost identical to the one in our non-Composer upstreams (down the road we can, hopefully, pull this file out of the other WP upstreams so there's a single source, and place into each repo using automation).

A comment has been added to `wp-config.php` instructing customers to add their configuration customizations there.

@greg-1-anderson and I also discussed including the [environment-specific files](https://github.com/roots/bedrock/tree/master/config/environments), but removing these files removes potential spots for merge conflicts for our customers. My assumption is customers would be editing these files on their own, and they are updated relatively frequently (2x last year) in roots/bedrock.

config/application.php
jspellman814: Added a note about those files here: https://github.com/pantheon-systems/wordpress-composer-managed/pull/9#issuecomment-1192004544

jazzsequence (approved):
Changes lgtm, just one question about possible duplication.

I will say that locally, while it works in Lando without a hitch, WP-CLI throws an error that `WP_HOME` isn't defined.

I can also confirm that subdomain multisite is working with these changes and the current nginx config.

web/wp-config-pantheon.php
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Don't show deprecations; useful under PHP 5.5
+error_reporting(E_ALL ^ E_DEPRECATED);
+/** Define appropriate location for default tmp directory on Pantheon */
+define('WP_TEMP_DIR', sys_get_temp_dir());
+
+// FS writes aren't permitted in test or live, so we should let WordPress know to disable relevant UI
+if (in_array($_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' )) && ! env('DISALLOW_FILE_MODS')) {
+    Config::define('DISALLOW_FILE_MODS', true);
```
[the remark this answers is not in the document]
jspellman814: This shouldn't ever differ, and I think it'd be safe to remove from this file. Including it was more part of practice to see how this `wp-config-pantheon` file differs from our default.

namespacebrian:
[Changes from this PR](https://github.com/pantheon-systems/wordpress-composer-managed/commit/bc11ddbd6529f7bb455587af07a1cfea2b9cfd5d#diff-53ea5a45a4b413ca49c7fac3f022e1dcbc2c203dcab4a1e7a34d1a8d65b26977L51-L56) were mentioned in https://github.com/pantheon-systems/wordpress-composer-managed/issues/25

PR #11, opened 2022-07-28
[CMS-833] [CMS-837] Move files from wordpress-bedrock-integrations into upstream

jspellman814:
This PR removes the need for [pantheon-systems/wordpress-bedrock-integrations](https://github.com/pantheon-systems/wordpress-bedrock-integrations) by moving the files from that repo into this upstream, as well as removes the [pantheon-systems/composer-scaffold](https://github.com/pantheon-systems/composer-scaffold) dependency (a fork of drupal/core-composer-scaffold that we'll no longer need to maintain).


composer.json
```diff
@@ -41,9 +41,11 @@
     "roots/wp-config": "1.0.0",
     "roots/wp-password-bcrypt": "1.1.0",
     "wpackagist-theme/twentytwentytwo": "^1.2",
-    "pantheon-systems/wordpress-bedrock-integrations": "^2.1",
     "pantheon-systems/pantheon-mu-plugin": "^1.0",
-    "pantheon-systems/composer-scaffold": "^1.0"
+    "pantheon-systems/quicksilver-pushback": "^2.1",
```
greg-1-anderson, line 45: We don't want to put quicksilver-pushback back into the upstream.

web/app/mu-plugins/loader.php
```diff
@@ -0,0 +1,90 @@
...
+ * Version: 1.0
+ * License: MIT
+ */
+
+if ( ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) ) {
+	return;
+}
+
+// Add your mu-plugins here.
+$mu_plugins = [
```
greg-1-anderson, line 16: If someone makes a custom upstream derived from this upstream, what are the odds that they might want to add an mu-plugin? Does anyone ever add an mu-plugin to pantheon-systems/WordPress? Do they have to modify this file?

If we put this file in the upstream, we need a strategy for how users will modify it (if necessary) without causing conflicts (e.g. if Pantheon later adds another mu-plugin). We have the same concern if this file is scaffolded, except the solution might be even more difficult in that instance.

web/wp-cli.yml
```diff
@@ -0,0 +1 @@
+path: wp
```
greg-1-anderson, line 1: This path needs to align with the folder structure defined in this repo, so it makes a lot of sense to commit it here rather than scaffold it from another project.

web/app/mu-plugins/loader.php
jspellman814: I'd say it's reasonably common for engineers to add mu-plugins, and I'm sure customers are doing that while using the `pantheon-systems/WordPress` upstream. They _don't_ need to modify this file - WordPress looks for PHP files right inside the mu-plugins directory, and will load those automatically (meaning there can be multiple loader files in this directory).

We could rename this to `pantheon-loader.php` (or something along those lines) and add a comment at the top instructing customers not to modify. Customers can then add their own PHP loader file to `web/app/mu-plugins` to load whichever mu-plugins they add to their site.

web/app/mu-plugins/pantheon-loader.php
```diff
@@ -0,0 +1,98 @@
...
+ * file inside the mu-plugins directory. This file may change in 
+ * future releases and modifications would cause conflicts when attempting
+ * to apply upstream updates.
+ */
+
+if ( ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) ) {
+	return;
+}
+
+// Add your mu-plugins here.
```
jspellman814, line 23: Remove comment

web/app/mu-plugins/pantheon-loader.php
```diff
@@ -0,0 +1,98 @@
...
+	if ( file_exists( $path ) ) {
+		include_once $path;
+	} else {
+		error_log( sprintf( 'Tried to load %s but the file was not found.', $file ) );
+	}
+}
+unset( $file );
+
+add_action(
+	'pre_current_active_plugins', function () use ( $mu_plugins ) {
```
jspellman814, line 41: Where is this from?

jspellman814: As far as I can tell, the origin of this code and the filter below (`network_admin_plugin_action_links`) is here: https://github.com/pantheon-systems/wp13n/commit/7bd722d2f8282db0258cbc2ae435c5a48274d2f7

And was later moved to [wordpress-bedrock-recommended](https://github.com/pantheon-systems/wordpress-bedrock-recommended/commit/66a621cba4ee20cecd51bc71f982590124d4e9b2#diff-d9815a88be153f1989fd34f7cb7841cbc844d1ac1e67769ac24318555922f991) and finally to this prototype. These filters are used to render the mu-plugins in the site plugin list at `wp/wp-admin/plugins.php?plugin_status=mustuse`, and I don't see any issue with including this.

web/app/mu-plugins/pantheon-loader.php
```diff
@@ -1,97 +0,0 @@
-<?php
```
jspellman814: Did some additional testing here (since creating CMS-837) and learned that `roots/bedrock-autoloader` [loads](https://github.com/roots/bedrock-autoloader/blob/master/src/Autoloader.php#L57) any plugin added to the `mu-plugins` directory. It also (like this file did) [displays](https://github.com/roots/bedrock-autoloader/blob/master/src/Autoloader.php#L77) the autoloaded plugins in the WP admin.

web/app/mu-plugins/loader.php
jazzsequence: If we're keeping this file as a `pantheon-loader` then we should provide a suggestion about how they can add their own loader file. Since this file was built to be generic and widely used, I'd say the comment instructing users to _not use this file_ tells them that they can copy the file as the base for their own loader file.

web/app/mu-plugins/pantheon-loader.php
jazzsequence: This loader file is a variation of the one we used at Human Made, which looks something like this (note, this is an older version, but it's probably still pretty close): https://github.com/humanmade/hm-base/blob/master/content/mu-plugins/loader.php

web/app/mu-plugins/pantheon-loader.php
jazzsequence: I'm glad you tested that. I remember seeing something in Bedrock docs that said something about that but I hadn't gotten around to validating or testing it.

jazzsequence (approved):
LGTM 👍

web/wp-cli.yml
sleddd: Hi, I am wondering, can you tell me why this file was never removed? There is already one that exists mapped to web/wp in the main directory? Bedrock does not have it either. Do you know why you left it in? And if you do not need it, can it be removed?

PR #12, opened 2022-07-29
[CMS-850] Add upstream-configuration directory

jspellman814:


CdrMarks (approved):

upstream-configuration/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "pantheon-upstreams/upstream-configuration",
+    "type": "project",
+    "version": "1.0.0",
```
greg-1-anderson, line 4: Using `version 1.0.0` can cause unexpected results. The requirements in the `require` section of this repo will be cached in the top-level composer.lock file. If the contents of the `require` section in this file change, then the composer.lock will not change unless the version here is also bumped.

jspellman814: I was looking at wordpress-project (which has a version defined in [this file](https://github.com/pantheon-upstreams/wordpress-project/blob/master/upstream-config/composer.json)), but noticing [drupal-composer-managed](https://github.com/pantheon-upstreams/drupal-composer-managed/blob/main/upstream-configuration/composer.json) doesn't contain a version. I'll update to match the latter.

greg-1-anderson: The drupal-composer-managed upstream solves this with a Composer script:

https://github.com/pantheon-systems/drupal-composer-managed/blob/default/upstream-configuration/scripts/ComposerScripts.php#L100

However, this breaks `--dry-run` mode in `composer update`, which disables Composer scripts. I tried using "version": "dev-main" here instead of the Composer script, but it did not work.

I recommend using the Composer script. It's not perfect, but I think it's the best of the options available.

greg-1-anderson (changes requested):
At a minimum, needs to include the updatesm-configuration path repository:

https://github.com/pantheon-systems/drupal-composer-managed/blob/default/composer.json#L17

Also consider whether we want the same Composer scripts we have in drupal-composer-managed.

greg-1-anderson (approved):

PR #24, opened 2022-09-19
[CMS-956] Updated gitignore

jazzsequence:
Copies and updates missing pieces from the [default WordPress upstream gitignore](https://github.com/pantheon-systems/WordPress/blob/default/.gitignore) into Composer-managed gitignore.


jspellman814 (approved):
LGTM 💥

PR #28, opened 2022-09-23
Update the Readme forking notes

jazzsequence:
The platform does not support `main` for custom upstreams. While this can be changed by Pantheors in the admin dashboard, that is not possible and adds a point of friction to users. Instead, as much as it pains us, we need to ask users to create a `master` branch and set that as the default for their custom upstreams, at least for the time being.


CdrMarks (approved):

jspellman814:
For custom upstreams, do we want to mention that customers can write their own CI to push to `master` from another branch if they want to track a more inclusively named default branch locally/internally?

PR #32, opened 2022-10-17
[BUGS-5424] Load .env file even if .env.local is not present.

### LINKED ISSUE #31: .env not loading without .env.local
Based on the current logic of selecting the appropriate env file, the `.env` is not getting loaded unless I _also_ have a `.env.local`.

Line 36 here does not include `.env`:
https://github.com/pantheon-systems/wordpress-composer-managed/blob/3318e7eb07b996b69ea4da801808092bd6225a05/config/application.php#L34-L36

So when `load()` is called on line 40, `$env_files` contains only `.env.pantheon`, even though the conditional _is_ checking that `.env` exists prior to loading:
https://github.com/pantheon-systems/wordpress-composer-managed/blob/3318e7eb07b996b69ea4da801808092bd6225a05/config/application.php#L38-L40

## Expected

`.env` should load whether or not `.env.local` exists.

## Actual
`.env` is only loaded if `.env.local` exists.

## Solution
This is likely as simple as adding `.env` at index 0 in the array on line 36. That would more closely match how Bedrock handles it:

https://github.com/roots/bedrock/blob/master/config/application.php#L32-L43

I'm happy to submit a PR! just wanted to open a discussion first.

kporras07:
Fixes #31

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -33,7 +33,7 @@ $webroot_dir = $root_dir . '/web';
  */
 $env_files = file_exists($root_dir . '/.env.local')
     ? ['.env', '.env.pantheon', '.env.local']
-    : ['.env.pantheon'];
+    : ['.env', '.env.pantheon'];
 
 $dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
 if (file_exists($root_dir . '/.env')) {


jazzsequence (approved):
🚢

PR #36, opened 2022-10-18
Pull the latest for all packages

jspellman814:
Prevent the need for future updates to composer.json file to avoid merge conflicts.


composer.json
```diff
@@ -56,9 +56,9 @@
     "oscarotero/env": "^2.1",
     "roots/bedrock-autoloader": "^1.0",
     "roots/bedrock-disallow-indexing": "^2.0",
-    "roots/wordpress": "6.0.3",
-    "roots/wp-config": "1.0.0",
-    "roots/wp-password-bcrypt": "1.1.0",
+    "roots/wordpress": "*",
```
kporras07, line 59: I'm not comfortable with "*" as a pretty messed up composer conflict could even downgrade the site instead of upgrading it.

Could you please put "^" in front of current version? If we ever need a major version change there in the future, hopefully at that time we'll have start states decoupled and we'll be able to do so.

composer.json
```diff
@@ -56,9 +56,9 @@
     "oscarotero/env": "^2.1",
     "roots/bedrock-autoloader": "^1.0",
     "roots/bedrock-disallow-indexing": "^2.0",
-    "roots/wordpress": "6.0.3",
-    "roots/wp-config": "1.0.0",
-    "roots/wp-password-bcrypt": "1.1.0",
+    "roots/wordpress": "^6.0.3",
```
jazzsequence, line 59: setting this to `"^6.0.3"` will mean that we need to manually update the `composer.json` again for WP 6.1 in a couple weeks. I don't know that there's a reason to not just let WordPress updates happen and do this instead since `"prefer-stable"` is true:

```suggestion
    "roots/wordpress": "*",
```

jspellman814: This is true too - thoughts on @kporras07 [concerns](https://github.com/pantheon-systems/wordpress-composer-managed/pull/36#discussion_r998424049)? Specifically:

> I'm not comfortable with "*" as a pretty messed up composer conflict could even downgrade the site instead of upgrading it.

kporras07: ^6.0.3 means everything between 6.0.3 and 7.0.0 (not included).

~6.0.3 means that @jazzsequence said.

Thanks composer for being confusing 😬

CdrMarks: Let's go with `^6.0.3` for now so we unblock John.  

@jspellman814 can you start a thread in our squad channel around best practices here versus when to use a `^` or `*`?  A good outcome from the conversation would be to add our decision as a page in https://github.com/pantheon-systems/Standards.

jazzsequence (approved):

PR #38, opened 2022-10-20
Improved debug log path and comment.

saqibsarwar:
Improved debug log path and comments to make it more understandable.


kporras07 (approved):
Thanks!

PR #39, opened 2022-10-21
[BUGS-5326] Add decoupled-release-pointer tag

jspellman814:
After testing, I'm fairly certain the reason our [decoupled upstream](https://github.com/pantheon-upstreams/decoupled-wordpress-composer-managed) isn't receiving updates is because the `deploy-public-upstream` job uses the same `release-pointer` tag in both repo steps. At the end of the `Copy commits to destination repo` step, the `release-pointer` tag is updated, and then the `Run decoupledpatch.sh and copy commits to decoupled destination repo` step runs and checks against the updated tag (always resulting in `No new commits found to release`).

This PR points the decoupled step towards its own tag, `decoupled-release-pointer`, which matches the most recent commit in the [upstream](https://github.com/pantheon-upstreams/decoupled-wordpress-composer-managed/commits/main).

Additionally, the devops directory does not exist in the decoupled upstream and we need to copy a few of these files to a temp directory before checking out that repo ([related test failure](https://app.circleci.com/pipelines/github/pantheon-systems/wordpress-composer-managed/168/workflows/6da88cb4-f32b-4af0-9295-f56f50050cc0/jobs/191)).

### DIFF
--- a/devops/scripts/deploy-decoupled-upstream.sh
+++ b/devops/scripts/deploy-decoupled-upstream.sh
@@ -20,8 +20,8 @@ echo "Preparing to release to upstream org"
 echo "-----------------------------------------------------------------------"
 echo
 
-# List commits between release-pointer and HEAD, in reverse
-newcommits=$(git log release-pointer..HEAD --reverse --pretty=format:"%h")
+# List commits between decoupled-release-pointer and HEAD, in reverse
+newcommits=$(git log decoupled-release-pointer..HEAD --reverse --pretty=format:"%h")
 commits=()
 
 # Identify commits that should be released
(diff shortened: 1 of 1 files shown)


kporras07 (dismissed):
I'm guessing you need to manually create and push the tag before merging, right? Other than that, LGTM

jspellman814:
> I'm guessing you need to manually create and push the tag before merging, right? Other than that, LGTM

Correct! Created here: https://github.com/pantheon-systems/wordpress-composer-managed/releases/tag/decoupled-release-pointer

namespacebrian (approved):
:rubber-stamp:

but you might want to get an approval from @jazzsequence / @CdrMarks since they have a better understanding of the interactions between these upstreams than I do

PR #44, opened 2022-11-02
[BUGS-5438] Move WP_ENVIRONMENT_TYPE set to application.php and use WP_ENV as fallback.

### LINKED ISSUE #37: WP_ENV value in .env is not initiating the correct WordPress environment.
I have to exclusively add `WP_ENVIRONMENT_TYPE='development'` to _.env_ file to enable the WordPress development environment.

I think `WP_ENV` value that already exists in _.env_ file should be used to set the WordPress environment.

https://github.com/pantheon-systems/wordpress-composer-managed/blob/4b558e3a249a79e4c4b7033ad61581b7f7285567/.env.example#L9

Thanks

kporras07:
closes #37


jspellman814 (commented):
Have a couple questions about this change that I've asked in the Jira card.

PR #46, opened 2022-11-04
Change from php_version to php-version

matt-bernhardt:
This repeats the change made by the Bedrock project in roots/bedrock#650 , making it so that the matrix of PHP versions is run as expected.


jspellman814 (approved):
LGTM. Thanks for the PR!

PR #47, opened 2022-11-07
[CMS-1106] Additional Lando configuration

jazzsequence:
This PR makes two changes for Lando local behavior and solves the issue described in CMS-1106 (local WordPress Composer Managed sites can't login on Lando).

* Adds an additional check for `.env.local` when a `lando` `PANTHEON_ENVIRONMENT` exists
* Updates the `.env.example` to use current Lando standards for the `.env` constant used for `WP_SITEURL` (see: https://docs.lando.dev/config/env.html#environment-files)


jspellman814 (approved):
:shipit:

PR #58, opened 2022-12-01
[CMS-1149] Remove dependabot

jspellman814:


jazzsequence (commented):
I feel hesitant about approving this without a Jira ticket that discusses/documents the reasons behind it. I _think_ this is okay, and even desirable, but I don't know what unintended side-effects there might be and there's no paper trail to follow to understand the reasons behind change.

jspellman814:
Sounds good - I will create a card for it. The main reason for removing it is that we don’t want to continue to make changes to the `composer.json` file (especially once this upstream is in GA), as that could introduce merge conflicts on customer's sites when clicking Apply Updates from the Pantheon dashboard.

I also don't know what unintended side-effects there might be.

jspellman814:
Tracking internally as CMS-1149

jazzsequence (approved):
Just wanted to make sure we had a paper trail for posterity. G2G 👍

PR #60, opened 2022-12-13
CMS-1112: Resolves issue where an unnecessary commit was being includ…

rwagner00:
…ed, resulting in merge conflicts.

This commit provides a way to manually add commit hashes to an exclude list, and adds the problem commit to that list.

The issue was caused by a desired and deliberate change to the commit type rules resulting in the deploy attempting to bring in a commit it had originally skipped, causing merge conflicts. There will likely be a small number of future instances where this is desirable, which has been facilitated by the array.

### DIFF
--- a/devops/scripts/deploy-decoupled-upstream.sh
+++ b/devops/scripts/deploy-decoupled-upstream.sh
@@ -29,10 +29,22 @@ echo
 newcommits=$(git log decoupled-release-pointer..HEAD --reverse --pretty=format:"%h")
 commits=()
 
+# There are a small number of commits which must be manually excluded, usually
+# due to changes in the commit type rules which would retroactively include commits
+# which were not initially included, causing merge conflicts. Commits which should
+# be excluded can be added to this array and will not be cherry picked.
+exclude_list=(0099a8b)
+
 # Identify commits that should be released
 for commit in $newcommits; do
   commit_type=$(identify_commit_type "$commit")
   if [[ $commit_type == "normal" ]] ; then
+
+    # Exclude commits which have been manually rejected
+    for item in "${exclude_list[@]}"; do
+      [[ $item == $commit ]] && echo "Commit has been manually excluded."; continue
+    done
+
     commits+=($commit)
   fi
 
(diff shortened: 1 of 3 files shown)


jazzsequence (commented):
Can you test this by triggering the deploy build step on this branch, too, (commenting out the actual deploy steps) so we can see if this works?

devops/scripts/deploy-decoupled-upstream.sh
```diff
@@ -88,19 +100,24 @@ cp /tmp/decoupled-README.md ./README.md
 git add .
 
 echo "Committing changes"
-git commit -F /tmp/commit_message --author='Pantheon Automation <bot@getpantheon.com>'
+# TODO: UNCOMMENT
+#git commit -F /tmp/commit_message --author='Pantheon Automation <bot@getpantheon.com>'
```
jazzsequence, line 104: Committing is probably fine if we aren't pushing. The environment should get destroyed between builds. 
🔢

devops/scripts/deploy-decoupled-upstream.sh
```diff
@@ -29,10 +29,22 @@ echo
 newcommits=$(git log decoupled-release-pointer..HEAD --reverse --pretty=format:"%h")
 commits=()
 
+# There are a small number of commits which must be manually excluded, usually
+# due to changes in the commit type rules which would retroactively include commits
+# which were not initially included, causing merge conflicts. Commits which should
+# be excluded can be added to this array and will not be cherry picked.
+exclude_list=(0099a8b)
```
jazzsequence, line 36:
```suggestion
exclude_list=("0099a8b")
```

devops/scripts/deploy-decoupled-upstream.sh
```diff
@@ -29,8 +29,19 @@ echo
...
+# due to changes in the commit type rules which would retroactively include commits
+# which were not initially included, causing merge conflicts. Commits which should
+# be excluded can be added to this array and will not be cherry picked.
+exclude_list=("0099a8b")
+
 # Identify commits that should be released
 for commit in $newcommits; do
+  # Exclude commits which have been manually rejected
+  for item in "${exclude_list[@]}"; do
+    [[ $item == $commit ]] && echo "Commit has been manually excluded."; continue
```
jazzsequence, line 42: Adding the commit hash here would be good and in line with other messages (e.g. `ommit c2e7468 contains both release and nonrelease changes. Skipping this commit.`)

```suggestion
    [[ $item == $commit ]] && echo "Commit ${commit} has been manually excluded.";
```

jazzsequence (approved):
Do not merge until after the break.

PR #66, opened 2023-01-24
DB-5248: add composer patches plugin

backlineint:
We'll be introducing an upstream dependency to decoupled-wordpress-composer-managed that uses a patch. To make this possible this PR adds and configures the composer patches plugin. We'll also need these changes to be syndicated to decoupled-wordpress-composer-managed when merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,8 @@
     "pantheon-upstreams/upstream-configuration": "dev-main",
     "wpackagist-plugin/lh-hsts": "^1.25",
     "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*"
+    "wpackagist-plugin/wp-native-php-sessions": "*",
+    "cweagans/composer-patches": "^1.7"
   },
   "require-dev": {
     "squizlabs/php_codesniffer": "^3.6.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -64,7 +64,8 @@
     "pantheon-upstreams/upstream-configuration": "dev-main",
     "wpackagist-plugin/lh-hsts": "^1.25",
     "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*"
+    "wpackagist-plugin/wp-native-php-sessions": "*",
+    "cweagans/composer-patches": "^1.7"
```
jazzsequence, line 68: Are you reasonably certain that the version of this dependency will never need to be higher than `1.x`?
If not, I would fall back to the precedent of using `*` for versioning since we're already requiring `stable` as our minimum stability. If we ever change this version in the future, it will lead to merge conflicts downstream.

backlineint: Never is not a favorite word of mine :) 

I think the risk of not using a constraint is that a breaking change could be released, which due to the nature of this plugin could result in new composer installs failing. 1.x has been stable for a long time, but you know... never say never.

Tough call, but I probably lean in the direction of using this constraint. That also follows the approach used by the Drupal, upstream: https://github.com/pantheon-systems/drupal-composer-managed/blob/default/composer.json#L22

jazzsequence: That seems reasonable. I am not _hugely_ concerned about the versioning (unless it's like a Composer v1 -> v2 sort of breaking change) especially as they don't push releases very frequently, but I wanted to at least call attention to it as a potential risk. This lgtm otherwise.

jazzsequence (dismissed):

jazzsequence (approved):

PR #70, opened 2023-02-03
Add CMS Platform to CODEOWNERS

jazzsequence:


jspellman814 (dismissed):

CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 # Code owners. See:
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-* @pantheon-systems/cms-ecosystem
\ No newline at end of file
+* @pantheon-systems/cms-ecosystem
+* @pantheon-systems/cms-platform
```
pwtyler, line 5:
```suggestion
* @pantheon-systems/cms-ecosystem @pantheon-systems/cms-platform
```

order takes precedence— if we only want cms-plat, it should be one name. if we want both teams notified, they should both be on the same line.

pwtyler (changes requested):

pwtyler (approved):

PR #79, opened 2023-03-24
[BUGS-6047] Force symbolic links in composer (no clone/embed folders).

### LINKED ISSUE #78: 'Failed to create symbolic link' - composer install attempts to create existing links when already run
## Background

The `post-install-cmd` attempts to **always** create symbolic links for the various 'root' files from `wp/*`, even when they already exist due to `composer install` previously being run (like on Local Development).

### Option 1 - Just force the sym-links

(Sending up a PR for this option.)

A tiny change of running `ln -sfn` will just force the sym-links destruction/recreation.  The `-n` here is needed so that the directories get replaced _too_, instead of embedded as children.

It'd have to be tested on OSX though, as I'm running WSL2.

#### Option 2a - Bash It!

This one would need two different changes:
1. A new script at `private/scripts/` named `bedrock-sym-links.sh`.
2. A update to run said script, instead of the existing command.

```bash
#!/usr/bin/env bash

function check_wp_symlinks() {
  core_file=web/wp-activate.php
  core_file2=web/wp-cron.php
  
  cd web
  echo -e "${yellow}Confirming or creating symlinks to the WordPress Core files.${normal}"
  if [ ! -L "$core_file" ]; then
    echo -e "${green}Links exist! Skipping!${normal}"
  else
    echo -e "${green}Creating symlinks${normal}"
    ln -s wp/* . || true
  fi
}

# Defines some global variables for colors.
normal=$(tput sgr0)
green=$(tput setaf 2)
yellow=$(tput setaf 3)

check_wp_symlinks
```

#### Option 2b - Composer It and abuse those Console Frameworks!

Yay for Composer. Yay for Symfony. Yay for Laravel! :D That said, maybe a Phase 2?

It would make the whole system more cross-platform compatible, because you'd just need to lean on some Symfony/Laravel components instead of hoping Bash is available..

- For knowledge, Symfony is likely the way to go, simply because you could use the `ScriptHandler.php` as a template, due to usage of both `Symfony\Component\Filesystem\Filesystem` and `Symfony\Component\Finder\Finder` from prior Pantheon CI/CD pipelines within the various build-kits used for CircleCI.
  - https://symfony.com/doc/current/components/filesystem.html#symlink
- For **WordPress**, Laravel is likely the way to go (if possible?), because this build-kit/workflow is aiming at getting both Sage **and** Acorn working on Pantheon, so might as well lean into it. :shrug: 
  - https://laravel.com/docs/10.x/filesystem#the-public-disk
  - https://laravel.com/api/10.x/Illuminate/Filesystem/Filesystem.html#method_link
  - **Genuinely** curious if the `StorageLinkCommand` class could be abused to pull this off, since the 'end result' is literally running PHP against `is_link()` and, if false, triggering PHP for `symlink()`....

cleverington:
Fixes #78 

Should **definitely** be tested within OSX.

Not sure if OSX has `-n` or not.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -114,7 +114,7 @@
       "php -r \"copy('.env.example', '.env');\""
     ],
     "post-install-cmd": [
-      "cd web; ln -s wp/* . || true",
+      "cd web; ln -sfn wp/* . || true",
       "rm web/wp-settings.php"
     ],
     "pre-update-cmd": [


jazzsequence:
I will give this a spin. If it works, I'm calling it good. 👍

jazzsequence (approved):
Tested this a bunch of times trying to break it. It didn't break. 😀

PR #80, opened 2023-03-24
[BUGS-6034] sage theme script adjustment

### LINKED ISSUE #72: composer install-sage fails when using a 'string' instead of dashes-underscores
## Scenario

I ran `composer install-sage` in order to crash-test the Sage theme on Pantheon.

Instead of telling it I wanted a theme along the lines of `feeling-sage` or just `sage`, I told the installer to create for me a theme named `Feeling Sage`. (Forgive the ADHD, I assumed the instructions were on 'what a theme is' and not 'what the folder-name is'! 😆 😆

**This failed by:**
1. Not having a folder named `Feeling`.
2. Trying to parse a version constraint named `Sage`.
3. Deleting the entire contents of `composer.json`, committing the change, and pushing to Pantheon! 😈 
4. Yay! It's all broken. Oh what a world.... haha.

### Fix the "Input" validation

On line 90 of `private/scripts/sage-theme-install.sh`, it runs validation against if the User says "**No**" or "no" or "N" or "n".

It would make more sense to validate **for** 'Yes', 'yes', etc.

### Option 1 - Reject Strings with Spaces
Though less ideal, at least rejecting the 'strings with spaces' means the **rest** of the script won't fail due to a mistyped 'Read' statement.

Would need to add some validation around line 78 of `private/scripts/sage-theme-install.sh` ( `read -p "Theme name: " -r sagename` ) which rejects and re-prompts the user to 'try again'.

Possibly a recursive function that runs until either the user quits or gets it right?

(Please pardon 'bad Bash'. Been a while since I've done up a bash script. lol.)

```bash
  read -p "Theme name: " -r sagename
  confirmThemeName "$sagename"
  

function confirmThemeName( ) {
  if [[ $sagename = *[[:space:]]* ]]
  then
      echo "Invalid theme name. Please avoid spaces."
      read -p "Theme name: " -r sagename
      confirmThemeName "$sagename"
  fi
}
```

### Option 2 - Allow Strings with Spaces (Recommended)
This would obviously be the best/smoothest solution, which would require a bit **more** validation on the parsing side, as you'd want to convert the spaces to dashes `-` or underscores `-`. (I say dashes, considering it is mostly a WordPress Best Practice.)

```bash
  read -p "Theme name: " -r sagename
  setThemeName "$sagename"
  

function setThemeName( ) {
  if [[ $sagename = *[[:space:]]* ]]
  then
      echo "Invalid theme name. Converting 'spaces' into dashes '-'."
      tr ' ' '-' < "$sagename"
  fi
}
```

(Could use this same approach via `tr '[A-Z]' '[a-z]' < "$sagename"` to convert 'uppercase' letters to 'lowercase', but that's probably a slippery slope as some themes might be OOO-declared enough to have the theme _itself_ as a Class.  I mean, **I** haven't seen one, but doesn't mean it isn't out there.)

cleverington:
Fixes #72 

**Two Changes:**
1. Updated the Restart trigger to **require** "Yes" or "y", etc. instead of "Not no".
2. Added a small 'checker' function to ensure the theme-name attempting to generate is both valid and useable.

Potential additional changes which could go into the new `confirmThemeName()` function:

```bash
  echo "Ensuring lower-space letters only."
  sagename=$(echo "$sagename" | tr '[:upper:]' '[:lower:]')

  echo "Ensuring theme-names are dashes instead of underscores."
  sagename=$(echo "$sagename" | tr '_' '-')
```

Didn't include them, as it felt like more of an internal decision for Pantheon? I know there are a **lot** of variables to be considered.

That said, this would just ensure the theme name **at creation** had no underscores ( `_` ), no spaces ( ` ` ), and was lowercase.

The only scenario I can think of where you _wouldn't_ want those things on a "Best Practice" kit is where the theme itself somehow exists as a stand-alone Class, but..... I can't mentally picture that? :shrug:

### DIFF
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -108,11 +108,22 @@ function get_info() {
 
 # Confirm user-submitted theme-name and ensure letters are lower-space
 function confirmThemeName() {
-  if [[ $sagename = *[[:space:]]* ]]
-  then
-      echo "Invalid theme name. Converting 'spaces' into dashes '-'."
-      sagename=$(echo "$sagename" | tr '[:space:]' '-')
-  fi
+  # Replace spaces with dashes and convert to lowercase
+  echo "Validating theme name..."
+  sagename=$(echo "$sagename" | tr '[:space:]' '-' | tr '[:upper:]' '[:lower:]')
+
+  # Replace underscores with dashes
+  sagename=${sagename//_/\-}
+
+  # Remove double dashes
+  while [[ $sagename == *--* ]]; do
+    sagename=${sagename/--/-}
+  done
+
+  # Remove trailing dash (if present)
+  sagename=${sagename%-}
+
+  echo "Theme name is: $sagename"
 }
 
 # Use terminus whoami to check if the user is logged in and exit the script if they are not.
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -111,7 +111,7 @@ function confirmThemeName() {
   if [[ $sagename = *[[:space:]]* ]]
   then
       echo "Invalid theme name. Converting 'spaces' into dashes '-'."
-      sagename=$(echo "$sagename" | tr '[ ]' '[-]')
+      sagename=$(echo "$sagename" | tr '[:space:]' '-')
   fi
 }
 
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -76,6 +76,7 @@ function get_info() {
   if [ -z "$sagename" ]; then
     echo -e "${yellow}Enter your theme name.${normal}\nThis is used to create the theme directory. As such, it should ideally be all lowercase with no spaces (hyphens or underscores recommended)\n"
     read -p "Theme name: " -r sagename
+    confirmThemeName "$sagename"
   else
     echo -e "${green}Theme name: ${sagename}${normal}"
   fi
@@ -87,7 +88,7 @@ function get_info() {
   SFTP hostname: ${sftphost}"
   read -p "Is this correct? (y/n) " -n 1 -r
   # If the user enters n, redo the prompts.
-  if [[ $REPLY =~ ^[Nn]$ ]]; then
+  if [[ ! $REPLY =~ ^[Yy]$ ]]; then
     echo -e "\nRestarting...\n"
 
     # Toggle the restarted state.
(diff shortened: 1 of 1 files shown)


jazzsequence:
@cleverington I tweaked the `confirmThemeName` function a bit. You wanna run a few tests and try to break it? It seems to be working well for me, I asked ChatGPT for help. 😅

cleverington (commented):
@jazzsequence I think we're probably skipping the scenario for, say, special characters and non-UTF-8 characters, but I think this solution addresses the issue at hand completely.

Also: Going _fancy_ there with that ChatGPT. 🥇 Did you tell it to only respond to "HAL"?  🗡️

jazzsequence:
That's a good catch! I tested a theme name using Kanji characters (thanks Google Translate) and it seemed to work. I _think_ UTF8 folder names should be okay, but my script is currently running, so I'll find out pretty quick if something breaks on Pantheon. In the meantime:

```
Theme name: このテーマ名を壊す
Validating theme name...
Theme name is: このテーマ名を壊す
You've entered:
  Site name: bugs-6034
  Theme name: このテーマ名を壊す
```
...and...
```
Checking if web/app/themes/このテーマ名を壊す is exists and if it's empty.

ls: web/app/themes/このテーマ名を壊す: No such file or directory
Installing Sage.
Creating a "roots/sage" project at "web/app/themes/このテーマ名を壊す
```

jazzsequence:
This is fun:
```
 create mode 100644 "web/app/themes/\343\201\223\343\201\256\343\203\206\343\203\274\343\203\236\345\220\215\343\202\222\345\243\212\343\201\231/.editorconfig"
```

jazzsequence:
eyyyy 🙌 

```
+--------------------+----------+--------+---------+
| name               | status   | update | version |
+--------------------+----------+--------+---------+
| twentytwentytwo    | active   | none   | 1.3     |
| このテーマ名を壊す | inactive | none   | 10.5.1  |
+--------------------+----------+--------+---------+
 [notice] Command: bugs-6034.dev -- wp theme list [Exit: 0]
Activating the このテーマ名を壊す theme.
 [warning] This environment is in read-only Git mode. If you want to make changes to the codebase of this site (e.g. updating modules or plugins), you will need to toggle into read/write SFTP mode first.
Success: Switched to 'Sage Starter Theme' theme.
 [notice] Command: bugs-6034.dev -- wp theme activate このテーマ名を壊す [Exit: 0]
```

jazzsequence (approved):

PR #81, opened 2023-03-24
[BUGS-6036] Setting script PHP version properly.

### LINKED ISSUE #74: Sage install script > pantheon.yml vs pantheon.upstream.yml
**Issue**:
When using the sage install script, PHP is updated to version 8 using the `pantheon.upstream.yml` file, but this only applies at the upstream level. 

**Resolution**:
Assuming users pull in this repo as the upstream, then the [script](https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/private/scripts/sage-theme-install.sh#L138) should instead install and add the `php_version` directive to the local repo's `pantheon.yml` as the change is not registered otherwise.

cleverington:
Fixes #74 

**Changes:**
1. Updating `pantheon.yml` instead of the `pantheon.upstream.yml`, which should really be untouched and/or manually updated.
2. Confirming that PHP 8.x (whether `8.0` **or** `8.1`/`8.2`) has not already been declared.
3. Confirming whether the PHP version has already been set and updating appropriately.

### DIFF
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -135,7 +135,23 @@ get_field() {
 # Update to PHP 8.0
 function update_php() {
   echo -e "\n\n${yellow}Updating PHP version to 8.0.${normal}"
-  sed -i '' "s/php_version: 7.4/php_version: 8.0/" pantheon.upstream.yml
+
+  # Testing for any version of PHP 8.x and/or PHP 7.4.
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):
This lgtm. If/when the Pantheon default PHP version is bumped to 8.0, we'll need to adjust this -- probably to specifically look for 7.x versions -- but this does what I wanted to accomplish with this fix. 🎉

PR #82, opened 2023-03-24
[BUGS-6035] create a new function to check jq

### LINKED ISSUE #73: composer install-sage fails and deletes composer.json (WSL2)
## Background
See #72, where I **thought** this issue was cropping up, but it is separate.

When running `composer install-sage`, one small part "soft fails" and continues building:

```bash
Attempting to add a post-install hook to composer.json.
Adding a post-install hook to composer.json.
./private/scripts/sage-theme-install.sh: line 239: jq: command not found
sed: can't read s,%sagedir%,web/app/themes/feeling-sage,: No such file or directory
[master f1d2aa7] [Sage Install] Add post-install-cmd hook to also run install on feeling-sage
 1 file changed, 136 deletions(-)
 rewrite composer.json (100%)
Already up to date.
```

## Issue / Bug
The part of the script within `/private/scripts/sage-theme-install.sh` is attempting to "rewrite" the existing `composer.json` with a small change to **also** run the `composer install` command within the "theme folder" on build (lines 237 - 242).

The problem is, it attempts to copy the existing `composer.json` to a new `composer.json.new` file using a non-existent command: `jq`.

```bash
./private/scripts/sage-theme-install.sh: line 239: jq: command not found
```

The `jq` command is not a 'default' command on the Ubuntu build for WSL2.  (Not sure if it is for Lando either. :shrug: )

### Option 1 - Rewrite the whole thing in Composer so its Platform Agnostic
Considering this is composer stuff doing composer things, it honestly would make more sense for `sage-theme-install.sh` to be a part of (and/or extend) the `\WordPressComposerManaged\ComposerScripts` class.

### Option 2 - Add WSL2 specific instructions
Add a caveat that using WSL2, users should either run the `composer install-sage` command using Lando (with `jq` installed) and/or install it themselves?

**Does** Lando come with it installed (considering both are Ubuntu)?

## Interim Solution

I'll just be adding the following to my `.lando.yml` file:

```yml
services:
  appserver:
    build_as_root:
      - curl -sL https://deb.nodesource.com/setup_16.x | bash -
      - apt-get install -y nodejs jq
```

That way `lando composer install-sage` will work for future projects.

Not going to test it immediately though, as I plan to just **manually** create the `post-install-cmd` hook-change.

jazzsequence:
if jq doesn't exist, look for brew or apt-get. if neither of those exist, exit hard (but provide a link to downloading jq and possible next steps).

This PR pulls in #83 which adds a `.lando.upstream.yml` to handle runs inside Lando.

Also adds shellcheck for linting the Sage Install script and runs it in GH actions.

**Note:** Shellchecking the scripts in [`devops/scripts`](https://github.com/pantheon-systems/wordpress-composer-managed/tree/default/devops/scripts) is tracked in CMSP-291 and out of scope for this PR.
Fixes #73

### DIFF
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -264,7 +264,7 @@ function check_jq() {
       echo '      "post-install-cmd": ['
       echo "          \"@composer install --no-dev --prefer-dist --ignore-platform-reqs --working-dir=web/app/themes/$sagename\""
       echo '      ],'
-      echo "${yellow}You might try running `lando composer install-sage` if you are running locally.${normal}"
+      echo "${yellow}You might try running `lando composer install-sage` if you are running locally. Alternately, you can try installing jq another way and then running the script again: https://stedolan.github.io/jq/download/${normal}"
       exit 1
     fi
   fi
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -310,7 +310,7 @@ function update_composer() {
   fi
 
   # Check for long-running workflows.
-  if [[ "$(terminus workflow:wait --max=1 "${sitename}".dev)" -eq *"running"* ]]; then
+  if [[ "$(terminus workflow:wait --max=1 "${sitename}".dev)" == *"running"* ]]; then
     echo -e "${yellow}Workflow still running, waiting another 30 seconds.${normal}"
     terminus workflow:wait --max=30 "$sitename".dev
   fi
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -264,7 +264,7 @@ function check_jq() {
       echo '      "post-install-cmd": ['
       echo "          \"@composer install --no-dev --prefer-dist --ignore-platform-reqs --working-dir=web/app/themes/$sagename\""
       echo '      ],'
-      echo "${yellow}You might try running `lando composer install-sage` if you are running locally. Alternately, you can try installing jq another way and then running the script again: https://stedolan.github.io/jq/download/${normal}"
+      echo "${yellow}You might try running 'lando composer install-sage' if you are running locally. Alternately, you can try installing jq another way and then running the script again: https://stedolan.github.io/jq/download/${normal}"
       exit 1
     fi
   fi
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -264,7 +264,7 @@ function check_jq() {
       echo '      "post-install-cmd": ['
       echo "          \"@composer install --no-dev --prefer-dist --ignore-platform-reqs --working-dir=web/app/themes/$sagename\""
       echo '      ],'
-      echo "${yellow}You might try running `lando composer install` if you are running locally.${normal}"
+      echo "${yellow}You might try running `lando composer install-sage` if you are running locally.${normal}"
       exit 1
     fi
   fi
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -149,7 +149,7 @@ get_field() {
   input="$(echo "$input" | sed -e '1d' -e '$d')"
   # $1: field name
   # $2: input string
-  echo "$2" | awk -v field="$1" '$1 -eq field { print $2 }'
+  echo "$2" | awk -v field="$1" '$1 == field { print $2 }'
 }
 
 # Update to PHP 8.0
--- a/private/scripts/sage-theme-install.sh
+++ b/private/scripts/sage-theme-install.sh
@@ -160,7 +160,7 @@ function update_php() {
   # Testing for any version of PHP 8.x and/or PHP 7.4.
   phpAlreadyVersion8=$(cat pantheon.yml | grep -c "php_version: 8.")
   phpDeclaredInFile=$(cat pantheon.yml | grep -c "php_version: 7.4")
-  
+
   # Only alter if not already PHP 8.x.
   if [ "$phpAlreadyVersion8" -eq 0 ]; then
     # Test for PHP version declartion already in pantheon.yml.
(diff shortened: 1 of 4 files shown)


jazzsequence:
@pwtyler I'd love for you to spot check my bash here.

private/scripts/sage-theme-install.sh
```diff
@@ -227,24 +226,49 @@ function add_symlink() {
...
     # Switch back to Git mode.
-    terminus connection:set $sitename.dev git
+    terminus connection:set "$sitename".dev git
 
   # Create the symlink to /files/cache.
-  cd web/app
+  cd web/app || return
   echo -e "${yellow}Adding a symlink to the cache directory.${normal}"
-  ln -sfn uploads/cache
+  ln -sfn uploads/cache .
```
pwtyler, line 240: Not knowing the scripts use super-well, can you speak more to each of these changes?

pwtyler (commented):
Overall bash script lgtm, a few questions/comments but no hard blocks.

private/scripts/sage-theme-install.sh
```diff
@@ -11,14 +10,14 @@ function get_info() {
   dashboard_link="https://dashboard.pantheon.io/sites/${id}#dev/code"
 
   # Unset the variables if we're doing this a second time.
-  if [ $is_restarted == 1 ]; then
+  if [ "$is_restarted" == 1 ]; then
```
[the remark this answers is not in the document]
pwtyler: As a matter of habit, if writing /bin/bash and not trying to be POSIX compatible,  be consistent about using `[[` and `==` across the script. I only use single `[` if I'm explicitly writing for POSIX shell.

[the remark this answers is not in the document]
jazzsequence: > You had some shorthand on one of my PRs for saying "this applies in multiple places" Insert that here.
🔢

private/scripts/sage-theme-install.sh
jazzsequence: shellcheck flagged them as errors. that's it.

private/scripts/sage-theme-install.sh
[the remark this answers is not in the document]
jazzsequence: Perhaps worth adding to CMSP-291

private/scripts/sage-theme-install.sh
```diff
@@ -144,23 +143,23 @@ function check_login() {
...
   # Remove leading and trailing whitespace from each line
-  input="$(echo "$input" | sed -e 's/^[ \t]*//')"
+  input="${input#"${input%%[![:space:]]*}"}"
 
   # Remove the first and last lines that are entirely dashes
   input="$(echo "$input" | sed -e '1d' -e '$d')"
   # $1: field name
   # $2: input string
-  echo "$2" | awk -v field="$1" '$1 == field { print $2 }'
+  echo "$2" | awk -v field="$1" '$1 -eq field { print $2 }'
```
pwtyler, line 152: Sorry for the confusion with my above comment. -eq is only for numeric comparisons. I think you want == here.
```suggestion
  echo "$2" | awk -v field="$1" '$1 == field { print $2 }'
```

private/scripts/sage-theme-install.sh
```diff
@@ -182,30 +181,30 @@ function install_sage() {
...
 
   # NPM the things
-  npm install --prefix $sagedir
-  npm run build --prefix $sagedir
+  npm install --prefix "$sagedir"
+  npm run build --prefix "$sagedir"
 
   # Remove /public from .gitignore
-  sed -i '' "s/\/public//" $sagedir/.gitignore
+  sed -i '' "s/\/public//" "$sagedir"/.gitignore
```
pwtyler, line 204: does this sed work on linux? I always forget how to write a cross-compatible version.

private/scripts/sage-theme-install.sh
```diff
@@ -227,24 +226,49 @@ function add_symlink() {
...
+      echo "${yellow}Attempted to install jq but could not find Brew (for MacOS) or apt-get (for WSL2/Linux). Exiting here. You'll need to add the following lines to your composer.json:${normal}"
+      echo '  "scripts": {'
+      echo '      "post-install-cmd": ['
+      echo "          \"@composer install --no-dev --prefer-dist --ignore-platform-reqs --working-dir=web/app/themes/$sagename\""
+      echo '      ],'
+      echo "${yellow}You might try running 'lando composer install-sage' if you are running locally. Alternately, you can try installing jq another way and then running the script again: https://stedolan.github.io/jq/download/${normal}"
+      exit 1
+    fi
+  fi
+}
```
pwtyler, line 270:
```suggestion
# Check if jq is installed. If it's not, try a couple ways of installing it.
function check_jq() {
  # Check if jq is already installed.
  if command -v jq &> /dev/null; then
      return # jq is already installed
  fi
  echo -e "${yellow}jq is not installed.${normal}"
  # Check if brew is installed and install jq with it
  if command -v brew &> /dev/null; then
    echo -e "${yellow}Installing jq with Homebrew.${normal}"
    brew install jq
    return
  fi
  # Check if apt-get is installed and install jq with it
  if command -v apt-get &> /dev/null; then
    echo -e "${yellow}Installing jq with apt-get.${normal}"
    sudo apt-get install -y jq
    return
  fi
  echo "${yellow}Attempted to install jq but could not find Brew (for MacOS) or apt-get (for WSL2/Linux). Exiting here. You'll need to add the following lines to your composer.json:${normal}"
  echo '  "scripts": {'
  echo '      "post-install-cmd": ['
  echo "          \"@composer install --no-dev --prefer-dist --ignore-platform-reqs --working-dir=web/app/themes/$sagename\""
  echo '      ],'
  echo "${yellow}You might try running 'lando composer install-sage' if you are running locally. Alternately, you can try installing jq another way and then running the script again: https://stedolan.github.io/jq/download/${normal}"
  exit 1
}
```

This is less bash and more how I prefer to avoid nesting as much as possible. feel free to discard.

pwtyler (changes requested):

pwtyler (dismissed):
went ahead and committed the one thing that _actually_ needed a fix. +1 unless you want to test and address either of my final comments.

private/scripts/sage-theme-install.sh
jazzsequence: I believe (anecdotally) that this script has been run inside `lando ssh` which would be Linux-based so I would tentatively say yes. I also think it shouldn't be a blocker because if it doesn't work, we'll find out.

jazzsequence:
@pwtyler I committed the suggestion that breaks the nested conditionals. I agree, fewer conditionals is better and that logic was much messier before. Ready for a final ✅

pwtyler (approved):
:shipit:

PR #83, opened 2023-03-24
[BUGS-6035] Adding upstream .lando file for installation of 'jq', 'npm', etc.

cleverington:
See #73 

@jazzsequence 

You can see via the `build_as_root` where its tossing the `apt-get` items to install.

My personal one always has `nano` on that list, because `vim` can perish.  #hillidieon

### DIFF
--- a/.lando.upstream.yml
+++ b/.lando.upstream.yml
@@ -1,0 +1,25 @@
+recipe: pantheon
+config:
+  framework: wordpress
+  xdebug: false
+
+events:
+  post-start:
+    - appserver: composer install
+
+services:
+  appserver:
+    build_as_root:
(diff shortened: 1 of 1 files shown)


jazzsequence:
🤔 If we can install `jq` via NPM...could we not just npm require it and run that binary?

Also: agree on `nano` > `vim` but I'm always in the minority 😅

jazzsequence:
Hmm....yeah, no, `jq` via NPM is jQuery...
https://www.npmjs.com/package/jq

There _is_ a Windows binary, though...
https://stedolan.github.io/jq/download/

jazzsequence (approved):

PR #86, opened 2023-03-29
[HOTFIX] Fix failing site builds.

jazzsequence:
This PR carves out an exception to the index.php file in post install.

Our post-install command was updated to create hard symlinks to the contents of the web/wp directory. We do not want this for the index.php file which is modified to point to /wp/


rwagner00 (approved):
LGTM

PR #90, opened 2023-03-30
Backport Bedrock updates

jazzsequence:
There's only really one relevant update to Roots/Bedrock that we don't address in some way or other in our fork, which is [this change](https://github.com/roots/bedrock/pull/658) to the ternary definition of the `WP_POST_REVISIONS` constant to a null coalesce.


CdrMarks (approved):
LGTM.

PR #97, opened 2023-05-17
[BUGS-6047]

jazzsequence:

PR #98, opened 2023-05-18
[BUGS-6047] Adds symlink check to avoid file exists error

jazzsequence:
This PR repurposes the Sage install script to be a multipurpose script to do different things. It takes parameters (currently to install sage or check whether we should create symlinks) and also takes a help param that displays usage info.

The symlink check script checks for each file that we would otherwise be creating a symlink to to see if it exists. If it does not, the symlink is created, and if it does, it is skipped.

Minor improvements and error handling has been added to the Sage install process as well.

This is a clean PR based out of #97 which had some non-release commits that changed the workflows in an attempt to get the builds to pass.


private/scripts/helpers.sh
```diff
@@ -1,5 +1,46 @@
...
+
+  # Check for a valid command.
+  if [ "$1" != "install_sage" ] && [ "$1" != "maybe_create_symlinks" ]; then
+    echo "${red}Invalid command specified.${normal}"
+    echo "${help_msg}"
+    exit 1;
+  fi
+
+  # Execute the command passed.
+  "$@"
```
pwtyler, line 41: would prefer to check that `$1` is a valid function name before executing. optional. Also, if none of the functions have arguments of their own, could use `$1` instead of `$@`

private/scripts/helpers.sh
```diff
@@ -1,5 +1,46 @@
...
+green=$(tput setaf 2)
+yellow=$(tput setaf 3)
+# Additional styles we aren't currently using.
+# italic=$(tput sitm)
+#blue=$(tput setaf 4)
+#magenta=$(tput setaf 5)
+#cyan=$(tput setaf 6)
+
+# Main function that runs the script.
+function main() {
```
pwtyler, line 16: nit: define main below functions it calls (style, not required)

private/scripts/helpers.sh
```diff
@@ -1,5 +1,46 @@
 #!/usr/bin/env bash
```
pwtyler, line 1: optional: prefer bash strict mode.
```
#!/bin/bash
set -euo pipefail
IFS=$'\n\t'
```

pwtyler (dismissed):
all of my suggestions are optional. LGTM when you're satisfied with testing.

private/scripts/helpers.sh
jazzsequence: In my mind, `main` is the thing that makes everything else go, and it's the point of interaction for the script, so to me, it makes more sense at the top because it's the thing that's interacted with first.

private/scripts/helpers.sh
jazzsequence: Can't exit on failures because some of those failures we need to see in the script (specifically the Sage install script) and exiting would break things that are otherwise handled gracefully.

private/scripts/helpers.sh
jazzsequence: >would prefer to check that $1 is a valid function name before executing

That's handled inside `main`: https://github.com/pantheon-systems/wordpress-composer-managed/blob/5bb5db58b4da9b2f137f5fea6bc95fa7a5822a69/private/scripts/helpers.sh#L22-L38

pwtyler: ah overlooked this

private/scripts/helpers.sh
pwtyler: Imo anything with the need to be handled gracefully should be wrapped in an if statement accordingly but fine as is.

pwtyler (approved):

PR #100, opened 2023-05-22
use $i instead of i

jazzsequence:
i is an unbound variable, $i checks the array value at index $i

### DIFF
--- a/devops/scripts/deploy-public-upstream.sh
+++ b/devops/scripts/deploy-public-upstream.sh
@@ -41,7 +41,7 @@ for commit in $newcommits; do
     delete=(${commit})
     for remove in "${delete[@]}"; do
       for i in "${commits[@]}"; do
-        if [ [ ${commits[$i]} = $remove ]]; then
+        if [[ ${commits[$i]} == "$remove" ]]; then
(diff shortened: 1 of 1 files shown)


jazzsequence:
ChatGPT seemed to think this was the right thing to do 😅

devops/scripts/deploy-public-upstream.sh
```diff
@@ -41,7 +41,7 @@ for commit in $newcommits; do
     delete=(${commit})
     for remove in "${delete[@]}"; do
       for i in "${commits[@]}"; do
-        if [ [ ${commits[i]} = $remove ]]; then
+        if [ [ ${commits[$i]} = $remove ]]; then
```
pwtyler, line 44: i fix lgtm.
```suggestion
        if [[ ${commits[$i]} == "$remove" ]]; then
```
If you're going to use bash [[, go ahead and double equal it. I'm not sure about all the quotes on $commits so left it alone but technically shellcheck would want that quoted too.

pwtyler (dismissed):

pwtyler (approved):

PR #101, opened 2023-05-22
Refactor how we handle "mixed" commits

jazzsequence:
Removes a bunch of failing code handling "mixed" commits that wasn't actually running except when it was failing and breaking things anyway.

"Mixed" commits fail the checks run on PRs, appropriately, and this code was never actually adding those commits to the array of commits to deploy. It was built to solve a problem where mixed commits would fail the deploy process, which we wanted to avoid. If the goal was "skipping" the commit, the way we can do that is just not adding it to the array of commits to deploy.

### DIFF
--- a/devops/scripts/deploy-public-upstream.sh
+++ b/devops/scripts/deploy-public-upstream.sh
@@ -39,7 +39,6 @@ for commit in $newcommits; do
   if [[ $commit_type == "mixed" ]] ; then
     2>&1 echo "Commit ${commit} contains both release and nonrelease changes. Skipping this commit."
     echo "You may wish to ensure that nothing in this commit is meant for release."
-    continue
   fi
 done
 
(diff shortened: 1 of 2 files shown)


devops/scripts/deploy-public-upstream.sh
```diff
@@ -33,19 +33,13 @@ for commit in $newcommits; do
...
     echo "You may wish to ensure that nothing in this commit is meant for release."
-    delete=(${commit})
-    for remove in "${delete[@]}"; do
-      for i in "${commits[@]}"; do
-        if [[ ${commits[$i]} == "$remove" ]]; then
-          unset 'commits[i]'
-        fi
-      done
-    done
+    continue
```
pwtyler, line 42:
```suggestion
```
with nothing below this if block, no continue necessary.

pwtyler (approved):

PR #105, opened 2023-06-23
[CMSP-481] Remove `upstream-require` script

jazzsequence:
This PR removes the `UpstreamRequire` method in `upstream-configuration/scripts/ComposerScripts.php` and adds the Composer hijacking magic ✨ that removes it from customer `composer.json` files without us creating merge conflicts.

**Note:** Before this upstream is officially GA, we should actually remove the `require` and `scripts-description` for `upstream-require` since, at that point, people should have already have updated and not have it anymore OR be brand new and should not be given something that's going to immediately be removed.

### DIFF
--- a/upstream-configuration/scripts/ComposerScripts.php
+++ b/upstream-configuration/scripts/ComposerScripts.php
@@ -15,37 +15,9 @@ class ComposerScripts
 {
    /**
     * Prepare for Composer to update dependencies.
-    *
-    * Composer will attempt to guess the version to use when evaluating
-    * dependencies for path repositories. This has the undesirable effect
-    * of producing different results in the composer.lock file depending on
-    * which branch was active when the update was executed. This can lead to
-    * unnecessary changes, and potentially merge conflicts when working with
-    * path repositories on Pantheon multidevs.
-    *
-    * To work around this problem, it is possible to define an environment
-    * variable that contains the version to use whenever Composer would normally
-    * "guess" the version from the git repository branch. We set this invariantly
-    * to "dev-main" so that the composer.lock file will not change if the same
-    * update is later ran on a different branch.
-    *
-    * @see https://github.com/composer/composer/blob/main/doc/articles/troubleshooting.md#dependencies-on-the-root-package
     */
     public static function preUpdate(Event $event)
-    {
-        $io = $event->getIO();
-
-        // We will only set the root version if it has not already been overriden
-        if (!getenv('COMPOSER_ROOT_VERSION')) {
-            // This is not an error; rather, we are writing to stderr.
-            $io->writeError("<info>Using version 'dev-main' for path repositories.</info>");
-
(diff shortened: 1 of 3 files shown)


upstream-configuration/scripts/ComposerScripts.php
```diff
@@ -97,34 +42,157 @@ public static function preUpdate(Event $event)
...
+        $platformPhpVersion = static::getCurrentPlatformPhp($event);
+        $pantheonPhpVersion = static::getPantheonPhpVersion($event);
+        $updatedPlatformPhpVersion = static::bestPhpPatchVersion($pantheonPhpVersion);
+        if ((substr($platformPhpVersion, 0, strlen($pantheonPhpVersion)) != $pantheonPhpVersion) && !empty($updatedPlatformPhpVersion)) {
+            $io->write("<info>Setting platform.php from '$platformPhpVersion' to '$updatedPlatformPhpVersion' to conform to pantheon php version.</info>");
+            $composerJson['config']['platform']['php'] = $updatedPlatformPhpVersion;
+        }
+
+        // Remove our upstream convenience scripts, if the user has not removed them.
+        if (isset($composerJson['scripts']['upstream-require'])) {
```
kporras07, line 93: This will be running these changes every time someone runs `composer update` (which includes `composer require`). Are you ok with this?

In DCM, we added a new post-update-hook and made sure to run this code only if that hook was absent: https://github.com/pantheon-systems/drupal-composer-managed/blob/default/upstream-configuration/scripts/ComposerScripts.php#L108

upstream-configuration/scripts/ComposerScripts.php
```diff
@@ -97,34 +42,157 @@ public static function preUpdate(Event $event)
...
+    /**
+     * postUpdate
+     *
+     * After "composer update" runs, we have the opportunity to do additional
+     * fixups to the project files.
+     *
+     * @param Composer\Script\Event $event
+     *   The Event object passed in from Composer
+     */
+    public static function postUpdate(Event $event)
```
kporras07, line 59: I think you're not using this anywhere

kporras07 (commented):
One question and one other comment.

Also, maybe we should add some docs about the composer-package? In DCM we added a small README: https://github.com/pantheon-systems/drupal-composer-managed/blob/default/upstream-configuration/README.md#upstream-configuration (it's ok if you consider that the composer suggestion is enough)

upstream-configuration/scripts/ComposerScripts.php
jazzsequence: Makes sense. Updated to include the same.

jazzsequence:
> Also, maybe we should add some docs about the composer-package?

We have that, too: https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/upstream-configuration/README.md

kporras07 (approved):
LGTM

PR #113, opened 2023-09-21
Run automated test of Sage Install Script

### LINKED ISSUE #112: Hey, we can actually test that our Sage Install Script runs
TIL that `macos-latest` and `windows-latest` exist as `runs-on` parameters in GitHub actions. 

We can (and should!) use these to validate our sage install script!

Caveats:
* the `macos-latest` image comes with everything we need. Which is great! but less great if we wanted to test the experience of someone, e.g. without Homebrew
* No idea what the Windows image is like, but it uses `windows-server` which might be different than desktop Windows (?)

jazzsequence:
This PR creates a new GH action workflow to test the Sage Theme install script (in `./private/scripts/helpers.sh`) on MacOS and Linux environments (`ubuntu-latest` and `macos-latest`). (Windows tests were attempted but ultimately abandoned due to issues in Terminus and a [lack of support for Windows Terminus installs](https://github.com/pantheon-systems/terminus#ubuntu--winwslubuntu).)

The matrix of tests has been configured to use PHP 8.x versions on each of the OSes it's being tested on (Roots Bedrock does not support < PHP 8).

In addition, changes to the script has been made so it can be run without user input via environment variables. These environment variables can be used to define any of the user input fields as well as whether the script is being run by a CI.

Future improvements could include Windows support (pending official support in the Terminus project), ~~an environment variable for PHP version (so each PHP version in the matrix is actually updated so the created multidev is using that PHP version)~~ (this has been rolled in) ~~and documentation for using the environment variables~~ (documentation has been added), and testing on WordPress multisite.

closes #112

### DIFF
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -21,7 +21,6 @@ jobs:
         with:
           github-token: ${{ github.token }}
           script: |
-            const core = require('@actions/core');
             const pr = context.payload.number;
             core.exportVariable('PR_NUMBER', pr);
       - name: Setup PHP
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -42,7 +42,7 @@ jobs:
           esac
 
           php_short=$(echo "${{ matrix.php-version }}" | tr -d '.')
-          pr_num=$(echo "$PR_NUMBER")
+          pr_num=$(echo ${{ steps.pr.outputs.pull_request_number }})
           echo "PR number: $pr_num"
           echo "OS: $os_short"
           echo "PHP: $php_short"
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -1,11 +1,13 @@
 name: Sage Install Tests
 on:
+  push:
+    paths:
+      - '.github/workflows/sage-test.yml'
   pull_request:
     branches:
       - main
     paths:
       - '/private/scripts/**'
-      - '/.github/workflows/sage-test.yml'
 jobs:
   test:
     name: Sage Install Tests
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -1,11 +1,11 @@
 name: Sage Install Tests
 on:
-  push:
   pull_request:
     branches:
       - main
     paths:
       - '/private/scripts/**'
+      - '/.github/workflows/sage-test.yml'
 jobs:
   test:
     name: Sage Install Tests
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -20,6 +20,13 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-version }}
+      - name: Get latest Terminus release
+        run: |
+          latest_version=$(curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | jq -r .tag_name)
+          echo "Terminus Latest version is $latest_version"
+          curl -L "https://github.com/pantheon-systems/terminus/releases/download/$latest_version/terminus.phar" -o terminus
+          chmod +x terminus
+          sudo mv terminus /usr/local/bin/
       - name: Install Composer Dependencies
         run: composer install --no-progress --no-suggest --prefer-dist
       - name: Run Sage Install Script
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        php-version: [8.0, 8.1, 8.2]
+        php-version: [8.0, 8.1, 8.2, 8.3]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -123,6 +123,7 @@ jobs:
           export SITENAME=wpcm-sage-install-tests
           export CI=1
           export SITEENV=$multidev_name
+          export PHPVERSION=${{ matrix.php-version }}
           cd ~/pantheon-local-copies/$SITENAME
           composer install-sage || exit_code=$?
           if [[ -n "${exit_code}" ]]; then
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -141,6 +141,6 @@ jobs:
         run: |
           cd ~/pantheon-local-copies/wpcm-sage-install-tests
           git fetch --tags origin
-          git tag -d pantheon_build_artifacts_$multidev_name
-          # Allow this to fail.
+          # Allow these to fail.
+          git tag -d pantheon_build_artifacts_$multidev_name || true
           git push origin --delete pantheon_build_artifacts_$multidev_name || true
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -55,6 +55,8 @@ jobs:
           echo "multidev_name=$multidev_name" >> $GITHUB_ENV
       - name: Get latest Terminus release
         run: |
+          latest_version=$(curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | jq -r .tag_name)
+          echo "Latest Terminus version: $latest_version"
           if [[ "${{ matrix.os }}" == "macos-latest" ]]; then
             echo "☕ Installing Terminus on macOS from Homebrew..."
             brew install pantheon-systems/external/terminus
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -26,16 +26,13 @@ jobs:
         uses: webfactory/ssh-agent@v0.8.0
         with:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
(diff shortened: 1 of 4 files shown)


.github/workflows/sage-test.yml
```diff
@@ -0,0 +1,149 @@
...
+        uses: actions/checkout@v3
+      - name: Install SSH keys
+        uses: webfactory/ssh-agent@v0.8.0
+        with:
+          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
+      - name: Get and Set PR number
+        id: pr
+        run: |
+          pr_num=$(gh pr view --json number -q .number || echo "")
+          echo "pr_num=$pr_num" >> $GITHUB_ENV
```
pwtyler, line 35: nit: env variables should be all caps

.github/workflows/sage-test.yml
```diff
@@ -0,0 +1,149 @@
...
+      - name: Get latest Terminus release
+        run: |
+          latest_version=$(curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | jq -r .tag_name)
+          echo "Latest Terminus version: $latest_version"
+          if [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            echo "☕ Installing Terminus on macOS from Homebrew..."
+            brew install pantheon-systems/external/terminus
+          else
+            echo "💻 Installing Terminus from phar on Linux..."
+            curl -L "https://github.com/pantheon-systems/terminus/releases/download/$latest_version/terminus.phar" -o terminus
```
pwtyler, line 74: github should have a latest release endpoint so you needn't look it up.

Optional.

.github/workflows/sage-test.yml
```diff
@@ -0,0 +1,149 @@
...
+          export PHPVERSION=${{ matrix.php-version }}
+          cd ~/pantheon-local-copies/$SITENAME
+          composer install-sage || exit_code=$?
+          if [[ -n "${exit_code}" ]]; then
+            echo "❌ Sage Install Script Failed with exit code ${exit_code}"
+            exit $exit_code
+          else
+            echo "✅ Sage Install Script passed!"
+            exit 0;
+          fi
```
pwtyler, line 138: why check the exit code? Won't CI bail if this exits non-zero?
```suggestion
          composer install-sage
          echo "✅ Sage Install Script passed!"
```

.github/workflows/sage-test.yml
```diff
@@ -0,0 +1,149 @@
...
+      - name: Delete multidev
+        if: always()
+        run: terminus multidev:delete --delete-branch --yes wpcm-sage-install-tests.$multidev_name
```
pwtyler, line 141: Is multidev env not useful to review in case of failed tests? Seems should should only run on success, and otherwise have validation _in front_ of the CI that ensures there are enough environments.

private/scripts/helpers.sh
```diff
@@ -54,6 +68,10 @@ function get_info() {
 
   # Unset the variables if we're doing this a second time.
   if [ "$is_restarted" -eq 1 ]; then
+    if [ "$is_ci" -eq 1 ]; then
```
pwtyler, line 71: `-eq` might be unhappy if compared to `""`? Default to 0 up on line 23, or test with -n/-z first or just use [[ == ]] instead.

private/scripts/helpers.sh
```diff
@@ -138,15 +156,16 @@ function get_info() {
...
+    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+      echo "Restarting..."
+
+      # Toggle the restarted state.
+      is_restarted=1
+      get_info
+    fi
   fi
 
   if [ -z "$sitename" ] || [ -z "$sagename" ] || [ -z "$sftpuser" ] || [ -z "$sftphost" ]; then
```
pwtyler, line 171: I can't comment on the line below (172) but `"echo "${red}Missing information!${normal} Make sure you've everything for all the prompts.""` needs a typo fix.

private/scripts/helpers.sh
```diff
@@ -230,13 +264,14 @@ function update_php() {
     # Test for PHP version declartion already in pantheon.yml.
     if [ ! "$phpDeclaredInFile" -eq 0 ]; then
       # Update version to 8.x.
-      sed -i '' "s/7.4/8.0/" pantheon.yml
+      sed -i '' "s/7.4/${phpversion}/" pantheon.yml
     else
       # Add full PHP version declaration to pantheon.yml.
-      echo "php_version: 8.0" >> pantheon.yml
+      echo "php_version: ${phpversion}" >> pantheon.yml
```
pwtyler, line 270:
```suggestion
      echo "" >> pantheon.yml
      echo "php_version: ${phpversion}" >> pantheon.yml
```
learned this the hard way when there's no newline at the end of the file...

private/scripts/helpers.sh
```diff
@@ -230,13 +264,14 @@ function update_php() {
...
     fi
-    git commit -am "[Sage Install] Update PHP version to 8.0"
-    git push origin master
+    git status
```
pwtyler, line 272:
```suggestion
```
still useful?

private/scripts/helpers.sh
```diff
@@ -359,47 +391,78 @@ function update_composer() {
...
+  if ! git push origin "$branch"; then
     echo "${red}Push failed. Stopping here.${normal}"
     echo "Next steps are to push the changes to the repo and then set the connection mode back to Git."
     exit 1;
   fi
 
   # Wait for the build to finish.
   echo "${yellow}Waiting for the deploy to finish.${normal}"
-  if ! terminus workflow:wait --max=90 "$sitename".dev; then
+  if [ "$is_ci" -eq 1 ]; then
```
pwtyler, line 424: same as above.

private/scripts/helpers.sh
```diff
@@ -359,47 +391,78 @@ function update_composer() {
...
+    echo "${yellow}Workflow still running, waiting another ${waittime} seconds.${normal}"
+    terminus workflow:wait --max="$waittime" "$sitename"."$siteenv"
   fi
+
+  git pull --ff --commit
 }
 
 # Finish up the Sage install process.
 function clean_up() {
+  # List the app/ehemes directory.
```
pwtyler, line 450:
```suggestion
  # List the app/themes directory.
```

private/scripts/helpers.sh
```diff
@@ -408,28 +471,32 @@ function clean_up() {
...
-  terminus connection:set "$sitename".dev sftp
+  terminus connection:set "$sitename"."$siteenv" sftp
 
-  # Open the site. This should generate requisite files on page load.
-  echo "${yellow}Opening the dev-${sitename}.pantheonsite.io to generate requisite files.${normal}"
-  open https://dev-"$sitename".pantheonsite.io
+  if [ "$is_ci" -ne 1 ]; then
+    # Open the site. This should generate requisite files on page load.
+    echo "${yellow}Opening the dev-${sitename}.pantheonsite.io to generate requisite files.${normal}"
+    open https://dev-"$sitename".pantheonsite.io
```
pwtyler, line 479: should this be `$siteenv` not dev?

pwtyler (changes requested):

.github/workflows/sage-test.yml
jazzsequence: I don't know that there's anything the multidev would tell you (that you couldn't get from workflow logs) that you wouldn't see in the output of the automated run. What's more, because the multidev environment is named based on the PR #, it will always create a multidev named `${os}-${php_version}-${pr_num}` and then we'd need to do all sorts of cleanup on the environment for subsequent runs on the same PR.

private/scripts/helpers.sh
jazzsequence: No because what I actually wanted was to see what version actually got added to the `pantheon.yml`. What I actually want is a `git diff`.

.github/workflows/sage-test.yml
jazzsequence: yeah, but if I do that, it wouldn't match with the other variables on line 55-57 (https://github.com/pantheon-systems/wordpress-composer-managed/pull/113/files#diff-2faa71adfe1b6b12a9b18a4d722c7463005f533d54912c2154589a1d6b91b3bcR55-R57) or I would need to do `pr_num=$(echo $PR_NUM)`

.github/workflows/sage-test.yml
```diff
@@ -0,0 +1,149 @@
...
+        run: |
+          cd ~/pantheon-local-copies/wpcm-sage-install-tests
+          composer install
+      - name: Run Sage Install Script
+        run: |
+          export SAGENAME=sage-test
+          export SITENAME=wpcm-sage-install-tests
+          export CI=1
+          export SITEENV=$multidev_name
+          export PHPVERSION=${{ matrix.php-version }}
```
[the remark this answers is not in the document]
jazzsequence: This works for everything except for `SITEENV` because the `GITHUB_ENV` variable `$multidev_name` doesn't get carried over into environment variables...maybe there's some ${{ github.env }} thing that could be used...

.github/workflows/sage-test.yml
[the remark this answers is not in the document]
jazzsequence: leaving because pulling the PR number directly from gh is simpler parsing out of GITHUB_REF

pwtyler (approved):

PR #114, opened 2024-04-05
Update automation

jazzsequence:
In order to make this repository public again, we needed to create a new Pantheon user with new secrets. While all of those things happened in the background with repository secrets, this PR makes some changes to the existing automations (including the Sage Theme Install test, which is why we needed a SSH key in the first place) to update and optimize them.

The `check commits` test has been moved to just run on pushes to `release` as part of the `deploy-public-upstream` workflow. 

Working Sage Install tests can be found here: https://github.com/pantheon-systems/wordpress-composer-managed/actions/runs/8576488930

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: Lint and Test
 
-on: [pull_request]
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ['7.4', '8.0', '8.1']
+        php-version: ['8.0', '8.1', '8.2', '8.3']
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
         php-version: ['7.4', '8.0', '8.1']
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
 
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -1,16 +1,16 @@
-name: CI
+name: Lint and Test
 
-on: [push]
+on: [pull_request]
```
pwtyler, line 3: This might need additional triggers if you want to act on GHA-created PRs, not sure what workflows are at play here.

.github/workflows/composer-diff.yml
```diff
@@ -0,0 +1,28 @@
+name: Composer Diff
+on:
+  pull_request:
+    paths:
+      - 'composer.lock'
+permissions:
+  contents: write
```
pwtyler, line 7:
```suggestion
```
Don't think this action is committing anything?

.github/workflows/sage-test.yml
```diff
@@ -1,4 +1,7 @@
 name: Sage Install Tests
+permissions:
+  pull-requests: read
+  contents: read
```
pwtyler, line 4: why these permissions needed?

private/scripts/helpers.sh
```diff
@@ -281,10 +281,14 @@ function update_php() {
 # Install sage and related dependencies.
 function install_sage_theme() {
   # Check if the directory $sagedir is empty. If it's not, bail.
-  echo "Checking if ${sagedir} is exists and if it's empty."
+  echo "Checking if ${sagedir} exists and if it's empty."
 
   if [ "$(ls -A "$sagedir")" ]; then
     echo "${red}Directory not empty!${normal}"
+    if [ "$is_ci" -eq 1 ]; then
```
pwtyler, line 288:
```suggestion
    if [[ "$is_ci" == 1 ]]; then
```
-eq will error if is_ci is an empty string. I don't see where it's coming from though so possibly this is unnecessary.

pwtyler (dismissed):
Approved with some potential adjustments. I may have missed something on the gha permissions.

.github/workflows/sage-test.yml
jazzsequence: because we're pulling down the repository and doing a bunch of stuff on it. the tests failed to run without the permissions.

.github/workflows/composer-diff.yml
jazzsequence: this was copy pasta from the last several composer-diff workflows. it might have been added with the idea that it's "writing" a comment to the PR but I don't think that's what this permission is for.

PR #115, opened 2024-04-08
[BUGS-7620] define WP_HOME

jazzsequence:
Adds a new `application.pantheon.php` file for Pantheon-specific modifications to `config/application.php`. Checks if `WP_HOME` is defined in `.env` and defines it if not.

### DIFF
--- a/config/application.pantheon.php
+++ b/config/application.pantheon.php
@@ -13,7 +13,7 @@
 use Roots\WPConfig\Config;
 use function Env\env;
 
-if ( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && 'lando' !== $_ENV['PANTHEON_ENVIRONMENT'] ) {
+if ( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && ! isset( $_ENV['LANDO'] ) ) {
 	// We can use PANTHEON_SITE_NAME here because it's safe to assume we're on a Pantheon environment if PANTHEON_ENVIRONMENT is set.
 	$sitename = $_ENV['PANTHEON_SITE_NAME'];
 	$baseurl = $_ENV['PANTHEON_ENVIRONMENT'] . '-' . $sitename . '.pantheonsite.io';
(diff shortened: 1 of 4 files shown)


config/application.pantheon.php
```diff
@@ -0,0 +1,40 @@
+<?php
+/**
+ * Pantheon platform application settings.
+ *
+ * IMPORTANT NOTE:
+ * Do not modify this file. This file is maintained by Pantheon.
```
greg-1-anderson, line 6: Usually, Pantheon-managed files in a composer-managed site are brought in via a dependency, e.g. pantheon-systems/drupal-integrations for Drupal sites. For WordPress composer-managed sites, this pattern was not followed because there were not many files to manage this way. Is there any appetite to revisit this decision?

config/application.pantheon.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Site-specific modifications belong in config/application.php, not this file.
+ * This file may change in future releases and modifications would cause
+ * conflicts when attempting to apply upstream updates.
+ */
+
+use Roots\WPConfig\Config;
+use function Env\env;
+
+if ( isset( $_ENV['PANTHEON_ENVIRONMENT'] ) && 'lando' !== $_ENV['PANTHEON_ENVIRONMENT'] ) {
```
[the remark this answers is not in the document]
jazzsequence: I will check to see if Lando has their own environment variables. We might be able to use something else. This is the same pattern we're using in https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/config/application.php#L77

config/application.pantheon.php
jazzsequence: Ultimately, I don't know that that would leave us in a better place than adding a file for Pantheon-maintained modifications as far as merge conflicts. Currently, we're making a one-line change in a part of `application.php` that is unlikely to change much, if they've changed it at all. If we added a dependency and moved our modifications there, we'd make changes to the `composer.json`, which users almost certainly _have_ made modifications to.

Also, environment-specific `application.php` files is a [recommended practice for Bedrock](https://github.com/roots/bedrock/blob/master/config/application.php#L143-L147). We're going against the grain a little bit by not using the `/environment/` directory, but since our changes apply to the platform and not a specific environment, this seemed to make the most sense. If people are familiar with Bedrock already, having a Pantheon-specific config file won't seem out of place.

I'm not broadly opposed to the idea, but this is also code that would otherwise have already existed in `wp-config.php` (or `wp-config-pantheon.php`) and I think it would be weirder to _not_ make it part of the core project.

rwagner00 (approved):

PR #121, opened 2024-04-30
Update Sage Test workflow

jazzsequence:
Resolves an issue where Sage Test workflow fails on multidev creation because `{os}-{phpver}-` (e.g. `mac-81-`) is not a valid multidev name.

Some additional changes were made in this workflow to ensure consistency and future debugging. The trigger was updated to `pull_request` from `push` so closing, reopening, or transitions could trigger the workflow. Additionally, the `pr_num` (which is a misnomer now) logic was updated to use the GHA run number instead of the PR number. This way the number is more unique and there is less change of multidev name collisions when subsequent pushes are made to the same PR.

An `Apply Upstream Updates` step was broken out of the workflow and set as a dependency for the Sage tests because it was previously being run inside the matrix of tests. This meant that any test that found upstream updates could fail other tests that are checking for upstream updates at the same time.

### DIFF
--- a/.github/workflows/sage-test.yml
+++ b/.github/workflows/sage-test.yml
@@ -70,8 +70,12 @@ jobs:
       - name: Get and Set PR number
         id: pr
         run: |
-          pr_num=${{ github.event.pull_request.number }} || "rel"
-          echo "PR_NUM=$pr_num" >> $GITHUB_ENV
+          pr_num=${{ github.event.pull_request.number }}
+          if [ "${pr_num}" == "" ]; then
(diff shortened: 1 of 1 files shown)


jazzsequence:
MacOS tests are failing on multidev creation with `Create a Multidev environment [error]  The default platform hostname mac-82-99-wpcm-sage-install-tests.pantheonsite.io already exists.` This is unrelated to the changes here and seems to be incidental (if the run # was 100, it would not fail, as indicated in the successful builds [here](https://github.com/pantheon-systems/wordpress-composer-managed/actions/runs/8901408304). This should be safe to merge.

PR #123, opened 2024-05-06
Use `terminus-github-actions` and Sage install script cleanup

jazzsequence:
* replaces (deprecated, duplicate) `pantheon-systems/action-install-terminus` with `pantheon-systems/terminus-github-actions`
* updates the output of the composer-diff comment
* generates a random theme name for the sage install (rather than just `sage-test`)
* adds a wait step before switching to sftp mode for the `add_symlink` step
* checks if the `/files/cache` directory exists before creating it
* updates multisite check logic to use `wp config is-true` (rather than running `theme enable` regardless)
* bails early if running in CI (last several steps are not needed for CI testing)

### DIFF
--- a/private/scripts/helpers.sh
+++ b/private/scripts/helpers.sh
@@ -459,8 +459,6 @@ function update_composer() {
 
 # Finish up the Sage install process.
 function clean_up() {
-  local is_multisite
-
   # List the app/themes directory.
   echo "${yellow}Checking the themes directory for ${sagename}.${normal}"
   # If the previous output did not include $sagename, bail.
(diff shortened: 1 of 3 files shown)


private/scripts/helpers.sh
```diff
@@ -452,6 +457,8 @@ function update_composer() {
 
 # Finish up the Sage install process.
 function clean_up() {
+  local is_multisite=""
```
pwtyler, line 460:
```suggestion
  local is_multisite
```
redundant

private/scripts/helpers.sh
```diff
@@ -460,8 +467,14 @@ function clean_up() {
     exit 1;
   fi
 
+  is_multisite=$(terminus wp -- "$sitename"."$siteenv" config get MULTISITE)
```
pwtyler, line 470: Prefer `config is-true`
https://developer.wordpress.org/cli/commands/config/is-true/

private/scripts/helpers.sh
```diff
@@ -331,11 +334,13 @@ function add_symlink() {
     mkdir web/app/uploads/cache
   fi
 
-  # Create a files/cache directory on the host.
-  sftp -P 2222 "$sftpuser"@"$sftphost" <<EOF
-    cd /files
-    mkdir cache
+  # Create a files/cache directory on the host if one does not already exist.
+  if [ "$(sftp -P 2222 "$sftpuser"@"$sftphost" <<< "ls /files" | grep -c "^cache$")" -eq 0 ]; then
```
pwtyler, line 338: would be nice if this is done with exit codes instead but maybe not worth the effort.

pwtyler (approved):
PR needs a description. Couple of minor cleanup suggestions at your discretion. Still ✅

private/scripts/helpers.sh
jazzsequence: I'm not sold on this yet as it still is getting a folder creation error in CI. Pretty sure this works if the script is run normally though and the script doesn't fail so ¯\_(ツ)_/¯

PR #124, opened 2024-05-07
[CMSP-132] shellcheck devops scripts

jazzsequence:
This pull request updates the linting Composer scripts in the project and adds a new `lint:devops` script to be used in automation and Pantheon engineers. It makes sure that only existing devops scripts are linted.

This PR also updates the scripts to shellcheck standards.

### DIFF
--- a/devops/scripts/check-commits.sh
+++ b/devops/scripts/check-commits.sh
@@ -9,9 +9,10 @@ set -euo pipefail
 . devops/scripts/commit-type.sh
 
 # List commits between release-pointer and HEAD, in reverse
+merge_base=""
 merge_base=$(git merge-base default HEAD)
+prcommits=""
 prcommits=$(git log "${merge_base}"..HEAD --pretty=format:"%h")
-
 status=0
 
 # Identify commits that should be released
--- a/devops/scripts/check-commits.sh
+++ b/devops/scripts/check-commits.sh
@@ -9,7 +9,8 @@ set -euo pipefail
 . devops/scripts/commit-type.sh
 
 # List commits between release-pointer and HEAD, in reverse
-prcommits=$(git log "$(git merge-base default HEAD)"..HEAD --pretty=format:"%h")
+merge_base=$(git merge-base default HEAD)
+prcommits=$(git log "${merge_base}"..HEAD --pretty=format:"%h")
 
 status=0
 
--- a/devops/scripts/check-commits.sh
+++ b/devops/scripts/check-commits.sh
@@ -9,7 +9,7 @@ set -euo pipefail
 . devops/scripts/commit-type.sh
 
 # List commits between release-pointer and HEAD, in reverse
-prcommits=$(git log $(git merge-base default HEAD)..HEAD --pretty=format:"%h")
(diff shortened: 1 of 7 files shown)


devops/scripts/check-commits.sh
```diff
@@ -9,7 +9,7 @@ set -euo pipefail
 . devops/scripts/commit-type.sh
 
 # List commits between release-pointer and HEAD, in reverse
-prcommits=$(git log $(git merge-base default HEAD)..HEAD --pretty=format:"%h")
+prcommits=$(git log "$(git merge-base default HEAD)"..HEAD --pretty=format:"%h")
```
pwtyler, line 12: Not a fan of nested subshells. set to a variable and pass to the next one

devops/scripts/commit-type.sh
```diff
@@ -40,7 +40,7 @@ function only_allowed_files() {
 
   affected_paths=$(git show "${commit}" --pretty=oneline --name-only | tail -n +2)
   for path in $affected_paths; do
-    if [[ " ${forbidden_files[*]} " =~ " ${path} " ]]; then
+    if [[ " ${forbidden_files[*]} " =~ ${path} ]]; then
```
pwtyler, line 43: I assume the linter told you to take the regex out of quotes. not sure either way. The spacing on either side is weird to begin with.

devops/scripts/deploy-decoupled-upstream.sh
```diff
@@ -101,6 +102,7 @@ for commit in "${commits[@]}"; do
 done
 
 echo "Executing decoupledpatch.sh"
+# shellcheck disable=SC1091
```
pwtyler, line 105: recommend source instead of disable.

devops/scripts/commit-type.sh
jazzsequence: This was the one I was most nervous about.

devops/scripts/deploy-decoupled-upstream.sh
jazzsequence: Disabling here because the file doesn't exist in the directory it's being sourced from in the script, so the linter is telling you "that file doesn't exist" (which is correct, but irrelevant).

devops/scripts/check-commits.sh
```diff
@@ -9,7 +9,8 @@ set -euo pipefail
 . devops/scripts/commit-type.sh
 
 # List commits between release-pointer and HEAD, in reverse
-prcommits=$(git log $(git merge-base default HEAD)..HEAD --pretty=format:"%h")
+merge_base=$(git merge-base default HEAD)
+prcommits=$(git log "${merge_base}"..HEAD --pretty=format:"%h")
```
pwtyler, line 14:
```suggestion
merge_base=""
merge_base=$(git merge-base default HEAD)
prcommits=""
prcommits=$(git log "${merge_base}"..HEAD --pretty=format:"%h")
```
Mostly calling out just because I would have expected this to have been called out by the linter, but it wasn't. 🤷🏻  Ususally doesn't like if you define a variable at the same time you assign its subshell value as that can mask errors in the subshell.

pwtyler (approved):
approving with one more optional todo

PR #125, opened 2024-06-04
[CMSP-150] Update wp-config filename to use `config/application.php`

jazzsequence:
This pull request updates the wp-config filename to use `config/application.php`. This change ensures that the correct filename is used on the WPMS network setup page.
Screenshot 2024-06-04 at 11 55 50 AM

### DIFF
--- a/web/app/mu-plugins/filters.php
+++ b/web/app/mu-plugins/filters.php
@@ -19,3 +19,13 @@ add_filter( 'pantheon.multisite.config_contents', function ( $config_contents ) 
 	$config_contents = str_replace( 'define(', 'Config::define(', $config_contents );
 	return $config_contents;
 } );
+
+/**
+ * Update the wp-config filename to use config/application.php.
+ *
+ * @return string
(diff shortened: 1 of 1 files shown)


web/app/mu-plugins/filters.php
```diff
@@ -19,3 +19,13 @@
...
 } );
+
+/**
+ * Update the wp-config filename to use config/application.php.
+ *
+ * @return string
+ */
+add_filter( 'pantheon.multisite.config_filename', function ( $config_filename ) {
+    $config_filename = 'config/application.php';
+    return $config_filename;
```
pwtyler, line 30:
```suggestion
    return 'config/application.php';
```
unnecessary assignment. optional.

pwtyler (approved):

web/app/mu-plugins/filters.php
jazzsequence: I was actually thinking about that just now, too

PR #128, opened 2024-06-11
fix: ensure managed wordpress files aren't omitted 

araphiel:
Fixes an issue where the following vcs controlled files were being ignored

The composer post install script is set to create symlinks from the `wp` to `web` folder, however the following files are not present in the original wp folder

```md
!web/wp-cli.yml
!web/wp-config.php
!web/wp-config-pantheon.php
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,8 @@ web/.htaccess
 # explicitly.
 web/license.txt
 web/readme.html
-web/index.php
 web/wp-*
+!web/index.php
 !web/wp-cli.yml
 !web/wp-config.php
 !web/wp-config-pantheon.php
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -24,8 +24,11 @@ web/.htaccess
 # explicitly.
 web/license.txt
 web/readme.html
-web/index.php
 web/wp-*
+!web/index.php
```
greg-1-anderson, line 28: `web/index.php` is placed by Composer, so it seems that this one, at a minimum, should remain ignored.

jazzsequence, line 28:
```suggestion
web/index.php
web/wp-*
```

jazzsequence: Agree 👍 
Updated to restore that line.

jazzsequence (approved):

PR #129, opened 2024-06-25
replace `tput` with a `safe_tput` function

jazzsequence:
This PR fixes a bug that appears in Workflow logs in the Pantheon dashboard.

Screenshot 2024-06-25 at 10 12 28 AM

The terminal environment used to render the workflow logs does not support `tput` and does not have a `$TERM` value, so setting the colors at the top of `helpers.sh` throws errors.

The `safe_tput` function allows `tput` to be used on systems that support it but does not error if `tput` is not available. In the cases when this is running (e.g. in Workflow logs) the colors are not really important and are only used when the symlinks are being created during `composer install`.

This PR also adds a `git push` after the Sage theme is built and installed because the script waits for an operation (e.g. `sync_code`) that isn't actually running. This is likely something that was removed but we still have the `workflow:wait` command in place.

### DIFF
--- a/private/scripts/helpers.sh
+++ b/private/scripts/helpers.sh
@@ -318,6 +318,7 @@ function install_sage_theme() {
   # Commit the theme
   git add "$sagedir"
   git commit -m "[Sage Install] Add the Sage theme ${sagename}."
+  git push
   echo "${green}Sage installed!${normal}"
 }
 
--- a/private/scripts/helpers.sh
+++ b/private/scripts/helpers.sh
@@ -2,17 +2,22 @@
 set -o pipefail
 IFS=$'\n\t'
 
-# Defines some global variables for colors.
-normal=$(tput sgr0)
-bold=$(tput bold)
-red=$(tput setaf 1)
-green=$(tput setaf 2)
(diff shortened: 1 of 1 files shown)


private/scripts/helpers.sh
```diff
@@ -2,17 +2,22 @@
...
-# Defines some global variables for colors.
-normal=$(tput sgr0)
-bold=$(tput bold)
-red=$(tput setaf 1)
-green=$(tput setaf 2)
-yellow=$(tput setaf 3)
+# Function to safely call tput
+safe_tput() {
+  tput "$@" 2>/dev/null || echo ""
+}
```
pwtyler, line 8: is the || echo necessary if we're not using `set -e`?

jazzsequence: to be fair, we might add `-e` later -- I have a ticket for it, I just need to figure out if I want to just close it out or do something with it. but can remove the `||` in the meantime.

pwtyler: if all its doing is preventing an error exit anyway, `|| true` should also work.

pwtyler (approved):

PR #130, opened 2024-06-25
[CMSP-1140] Fix core resource URLs for non-main sites in subdirectory multisite network

jazzsequence:
This pull request fixes the issue with core resource URLs for non-main sites in a subdirectory multisite network. Previously, the URLs were not being correctly adjusted, causing broken links to core resources (like CSS or JavaScript files). This PR adds a filter function `fix_core_resource_urls` that corrects the URLs by replacing the current site path with the main site URL. The filter is only applied to non-main sites in a subdirectory multisite network.

This problem comes from the custom directory structure used by Roots Bedrock. All the filter hooks are set to priority `9` so they can be easily overridden downstream.

### DIFF
--- a/web/app/mu-plugins/filters.php
+++ b/web/app/mu-plugins/filters.php
@@ -74,12 +74,12 @@ function fix_core_resource_urls( $url ) {
 
 // Only run the filter on non-main sites in a subdirectory multisite network.
 if ( is_multisite() && ! is_subdomain_install() && ! is_main_site() ) {
-    add_filter( 'script_loader_src', 'fix_core_resource_urls', 10 );
-    add_filter( 'style_loader_src', 'fix_core_resource_urls', 10 );
-    add_filter( 'plugins_url', 'fix_core_resource_urls', 10 );
-    add_filter( 'theme_file_uri', 'fix_core_resource_urls', 10 );
-    add_filter( 'stylesheet_directory_uri', 'fix_core_resource_urls', 10 );
-    add_filter( 'template_directory_uri', 'fix_core_resource_urls', 10 );
-    add_filter( 'site_url', 'fix_core_resource_urls', 10 );
-    add_filter( 'content_url', 'fix_core_resource_urls', 10 );
+    add_filter( 'script_loader_src', 'fix_core_resource_urls', 9 );
+    add_filter( 'style_loader_src', 'fix_core_resource_urls', 9 );
+    add_filter( 'plugins_url', 'fix_core_resource_urls', 9 );
+    add_filter( 'theme_file_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'stylesheet_directory_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'template_directory_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'site_url', 'fix_core_resource_urls', 9 );
+    add_filter( 'content_url', 'fix_core_resource_urls', 9 );
 }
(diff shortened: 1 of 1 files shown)


web/app/mu-plugins/filters.php
```diff
@@ -28,3 +28,58 @@
...
+    // Correct the path for core resources.
+    if ( strpos( $path, $current_site_path . 'wp-includes/' ) !== false ) {
+        $path = str_replace( $current_site_path . 'wp-includes/', 'wp-includes/', $path );
+    } elseif ( strpos( $path, $current_site_path . 'wp-admin/' ) !== false ) {
+        $path = str_replace( $current_site_path . 'wp-admin/', 'wp-admin/', $path );
+    } elseif ( strpos( $path, $current_site_path . 'wp-content/' ) !== false ) {
+        $path = str_replace( $current_site_path . 'wp-content/', 'wp-content/', $path );
+    } else {
+        return $url; // Return the original URL if no changes were needed.
+    }
```
pwtyler, line 62:
```suggestion
$core_paths = ['wp-includes/', 'wp-admin/', 'wp-content/'];
$path_modified = false;

foreach ($core_paths as $core_path) {
    if (strpos($path, $current_site_path . $core_path) !== false) {
        $path = str_replace($current_site_path . $core_path, $core_path, $path);
        $path_modified = true;
        break;
    }
}

// If the path was not modified, return the original URL.
if (!$path_modified) {
    return $url;
}
```

Not sure if we write test cases for these functions but this seems ripe for that.

web/app/mu-plugins/filters.php
```diff
@@ -28,3 +28,58 @@
...
+// Only run the filter on non-main sites in a subdirectory multisite network.
+if ( is_multisite() && ! is_subdomain_install() && ! is_main_site() ) {
+    add_filter( 'script_loader_src', 'fix_core_resource_urls', 9 );
+    add_filter( 'style_loader_src', 'fix_core_resource_urls', 9 );
+    add_filter( 'plugins_url', 'fix_core_resource_urls', 9 );
+    add_filter( 'theme_file_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'stylesheet_directory_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'template_directory_uri', 'fix_core_resource_urls', 9 );
+    add_filter( 'site_url', 'fix_core_resource_urls', 9 );
+    add_filter( 'content_url', 'fix_core_resource_urls', 9 );
```
pwtyler, line 84:
```suggestion
$filters = [
    'script_loader_src',
    'style_loader_src',
    'plugins_url',
    'theme_file_uri',
    'stylesheet_directory_uri',
    'template_directory_uri',
    'site_url',
    'content_url'
];
foreach ($filters as $filter) {
    add_filter($filter, 'fix_core_resource_urls', 9);
}
```

pwtyler (approved):
Both of my suggestions are code style so not blocking on either— my primary question I suppose is, is this something that could/should be contributed upstream to Bedrock, or is it specific to our implementation?

If we have a place for tests I would love to see this filter put through the wringer.

jazzsequence:
> is this something that could/should be contributed upstream to Bedrock, or is it specific to our implementation?

I'm not convinced it's Bedrock but I'm also not sure how it's not. It's possible they changed things and we haven't pulled down their latest code because this wasn't an issue a year ago. At any rate, I'm pretty confident that we're isolating to only scenarios that would actually be affected but I agree about tests. Currently, we don't have any tests, but I believe we have a ticket to create some, so...

web/app/mu-plugins/filters.php
jazzsequence:
```suggestion
$filters = [
    'script_loader_src',
    'style_loader_src',
    'plugins_url',
    'theme_file_uri',
    'stylesheet_directory_uri',
    'template_directory_uri',
    'site_url',
    'content_url'
];
foreach ($filters as $filter) {
    add_filter( $filter, 'fix_core_resource_urls', 9 );
}
```

PR #131, opened 2024-06-26
Update PHP version to 8.1 and other updates to `helpers.sh`

jazzsequence:
This PR updates the `helpers.sh` script's `update_php` function to default to a minimum of PHP 8.1 since Sage no longer supports 8.0.

The update_php function has therefore been updated to check for a minimum of 8.1 but _additionally_ updates have been made so that `update_php` can be called directly from the script outside of the sage install workflow, so we can test _that component_ separately. 

BATS tests have been created to test the `update_php` function for all versions from 7.4 - 8.3.

This PR also makes the following minor(ish) adjustments to the sage install script workflow:
- normalizes the generated theme names before the sage install script is called (fixes an issue where tests were failing because themes were created that included slashes in the directories that were created)
- silences the multisite check (sends the output of the check to `/dev/null` so a WP-CLI warning is not thrown if the `MULTISITE` constant does not exist)
- adds a check to see if the Sage theme exists before attempting to activate it

### DIFF
--- a/private/scripts/helpers.sh
+++ b/private/scripts/helpers.sh
@@ -484,7 +484,7 @@ function clean_up() {
   # Get the themes.
   themelist=$(terminus wp -- "$sitename"."$siteenv" theme list --format=csv | tr -d '\n' | tr -d ' ')
 
-  if ! echo "$themelist" | grep -q -E "(^|,)$sagename($|,)"; then
+  if ! echo "$themelist" | grep -q "$sagename"; then
     echo "${red}Theme $sagename not found in the theme list. Exiting here.${normal}"
     echo "$themelist"
     exit 1;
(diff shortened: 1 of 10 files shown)


private/scripts/helpers.sh
```diff
@@ -476,17 +476,24 @@ function clean_up() {
...
+    exit 1;
+  fi
 
   # Activate the new theme
   echo "${yellow}Activating the ${sagename} theme.${normal}"
   if ! terminus wp -- "$sitename"."$siteenv" theme activate "$sagename"; then
+    terminus wp -- "$sitename"."$siteenv" theme list
     echo "${red}Theme activation failed. Exiting here.${normal}"
     echo "Check the theme list above. If the theme you created is not listed, it's possible that the deploy has not completed. You can try again in a few minutes using the following command:"
     echo "terminus wp -- $sitename.dev theme activate $sagename"
```
pwtyler, line 499: Is this still accurate if we're exiting above in the case its not found

pwtyler (approved):

private/scripts/helpers.sh
jazzsequence: Probably not. 😁

PR #133, opened 2024-06-27
Bump PHP version to 8.2 and DB version to 10.6

jazzsequence:
This pull request updates the PHP version to 8.2 and the DB version to 10.6.

This is good to review, but I am asking in the Roots Bedrock Discord to make sure there aren't good reasons to _not_ upgrade PHP to 8.2 by default. (Sounds like it's a non-issue.)

This PR also bumps the required PHP version in `composer.json` to the version in the `pantheon.yml` file if the `require.php` version is less than the minimum required by Bedrock. This automates the process to bring (and keep) the `require.php` value in the `composer.json` in line with the version running on the platform and the Bedrock required versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
     }
   ],
   "require": {
-    "php": ">=8.0",
+    "php": ">=8.1",
     "composer/installers": "^2.2",
     "vlucas/phpdotenv": "^5.5",
     "oscarotero/env": "^2.1",
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
     }
   ],
   "require": {
-    "php": ">=8.1",
+    "php": ">=8.0",
     "composer/installers": "^2.2",
     "vlucas/phpdotenv": "^5.5",
     "oscarotero/env": "^2.1",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -50,7 +50,7 @@
     }
   ],
   "require": {
-    "php": ">=8.0",
+    "php": ">=8.1",
```
greg-1-anderson, line 53: This could potentially cause merge conflicts. If Bedrock requires PHP 8.1+, and since we require a couple of Bedrock dependencies, we should pick up their minimum PHP version automatically.

If it did turn out to be important to bump this for some reason, we could do it [at the same time we adjust the platform.php version](https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/upstream-configuration/scripts/ComposerScripts.php#L63), or at least in the same vicinity.

greg-1-anderson (dismissed):
Omit changes to composer.json file; update via scripts if necessary.

composer.json
jazzsequence:
```suggestion
    "php": ">=8.0",
```

pwtyler (approved):

PR #134, opened 2024-06-27
Bring WPCM in line with Bedrock v1.24.x

jazzsequence:
This pull request makes changes to bring the `application.php` and `composer.json` files in line with Bedrock v1.24.x.

Included in the change are updates the usage of Dotenv in the codebase to use the `createImmutable()` method instead of the deprecated `createUnsafeImmutable()` method. It also removes the unnecessary post install script that copies the .env.example file. This improves the security and maintainability of the codebase.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,9 +111,6 @@
       ]
   },
   "scripts": {
-    "post-root-package-install": [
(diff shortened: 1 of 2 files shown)


pwtyler (approved):
lgtm with tests passing

PR #137, opened 2024-06-28
Update WordPress URL handling and cookie settings

jazzsequence:
This pull request includes several changes to improve the handling of URLs in WordPress particularly for multisite. 

Included in the update are helper functions to check and strip double-slashes from URLs, namespacing, enforcing strict types, and a version bump for the filters mu-plugin. 

Importantly, this PR enforces the `/wp` prefix to admin and login URLs which prevents an infinite login redirect loop. 

The PR also adds a cookie domain fix for multisite logins which is a common issue for subdomain multisite.

### DIFF
--- a/web/app/mu-plugins/filters.php
+++ b/web/app/mu-plugins/filters.php
@@ -146,8 +146,6 @@ add_filter( 'admin_url', __NAMESPACE__ . '\\add_wp_prefix_to_login_and_admin_url
  * @return bool True if the URL is a login or admin URL. False if it's not or is not actually a URL.
  */
 function __is_login_url( string $url ) : bool {
-	$is_login_url = false;
-
 	// Validate that the string passed was actually a URL.
 	if ( ! preg_match( '/^https?:\/\//i', $url ) ) {
 		$url = 'http://' . ltrim( $url, '/' );
@@ -155,20 +153,15 @@ function __is_login_url( string $url ) : bool {
 
 	// Bail if the string is not a valid URL.
 	if ( ! wp_http_validate_url( $url ) ) {
-		return $is_login_url;
+		return false;
 	}
 
-	// Login page?
-	if ( strpos( $url, 'wp-login' ) !== false ) {
-		$is_login_url = true;
+	// Check if the URL is a login or admin page
+	if (strpos($url, 'wp-login') !== false || strpos($url, 'wp-admin') !== false) {
+		return true;
 	}
 
-	// Admin page?
-	if ( strpos( $url, 'wp-admin' ) !== false ) {
-		$is_login_url = true;
-	}
-
-	return $is_login_url;
+	return false
 }
 
 /**
--- a/web/app/mu-plugins/filters.php
+++ b/web/app/mu-plugins/filters.php
@@ -116,6 +116,14 @@ function adjust_main_site_urls( string $url ) : string {
 add_filter( 'home_url', __NAMESPACE__ . '\\adjust_main_site_urls', 9 );
 add_filter( 'site_url', __NAMESPACE__ . '\\adjust_main_site_urls', 9 );
 
+/**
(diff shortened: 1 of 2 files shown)


web/app/mu-plugins/filters.php
```diff
@@ -97,17 +101,96 @@ function fix_core_resource_urls( $url ) {
...
+		$is_login_url = true;
+	}
+
+	// Admin page?
+	if ( strpos( $url, 'wp-admin' ) !== false ) {
+		$is_login_url = true;
+	}
+
+	return $is_login_url;
+}
```
pwtyler, line 172: seems like we've almost got two patterns at play that got mixed— but looking at this, no need for `$is_login_url` to be a variable in the first place.
```suggestion
function __is_login_url( string $url ) : bool {
	// Validate that the string passed was actually a URL.
	if ( ! preg_match( '/^https?:\/\//i', $url ) ) {
		$url = 'http://' . ltrim( $url, '/' );
	}

	// Bail if the string is not a valid URL.
	if ( ! wp_http_validate_url( $url ) ) {
		return false;
	}

	// Check if the URL is a login or admin page
	if (strpos($url, 'wp-login') !== false || strpos($url, 'wp-admin') !== false) {
		return true;
	}

	return false
}
```

web/app/mu-plugins/filters.php
```diff
@@ -97,17 +101,96 @@ function fix_core_resource_urls( $url ) {
...
+ */
+function add_wp_prefix_to_login_and_admin_urls( string $url ) : string {
+	if ( __is_login_url( $url ) ) {
+		if ( strpos( $url, '/wp/' ) === false ) {
+			$url = __normalize_wp_url( preg_replace( '/(\/wp-(login|admin))/', '/wp/$1', $url ) );
+		}
+	}
+
+	return $url;
+}
```
pwtyler, line 135: can you guess what I'm going to say?
```suggestion
function add_wp_prefix_to_login_and_admin_urls( string $url ) : string {
	if (  ! __is_login_url( $url ) ) {
		return $url;
	}
	if ( strpos( $url, '/wp/' ) !== false ) {
		return $url;
	}
	return __normalize_wp_url( preg_replace( '/(\/wp-(login|admin))/', '/wp/$1', $url ) );
}
```

web/app/mu-plugins/filters.php
```diff
@@ -97,17 +101,96 @@ function fix_core_resource_urls( $url ) {
...
+		$remaining_url = substr( $url, $scheme_end_pos + 3 );
+	} else {
+		$scheme = '';
+		$remaining_url = $url;
+	}
+
+	// Normalize the URL to remove any double slashes.
+	$url = str_replace( '//', '/', $remaining_url );
+	return $scheme . $url;
+}
```
pwtyler, line 196: Feels dirty not to use parse_url() to get the scheme.
```suggestion
function __normalize_wp_url(string $url): string {
    $scheme = parse_url($url, PHP_URL_SCHEME);

    $scheme_with_separator = $scheme ? $scheme . '://' : '';

    // Remove the scheme from the URL if it exists.
    $remaining_url = $scheme ? substr($url, strlen($scheme_with_separator)) : $url;

    // Normalize the remaining URL to remove any double slashes.
    $normalized_url = str_replace('//', '/', $remaining_url);

    // Reconstruct and return the full normalized URL.
    return $scheme_with_separator . $normalized_url;
}
```
You'll laugh but I wrote this twice because I couldn't decide if I disliked ternary ifs or else statements more, and I settled on allowing ternaries today.

pwtyler (approved):
Sorry, this dropped off my radar. I have style options but nothing to withhold a ✅ on. All are at you discretion.

Feels wrong to introduce three new methods without tests but I understand we have follow-on work to add testing.

PR #138, opened 2024-07-08
[CMSP-1459] Fixes URL normalization process

jazzsequence:
This pull request 

* adds a new helper function that re-parses and normalizes a URL from its parts
* updates the existing WP URL normalization function to use the new helper function.
* pre-sets the endpoint option so the updated URL handling does not break existing sites expecting the endpoint to be at `/wp/graphql` while still allowing that option to be set via the plugin options.
* adds Playwright tests (adapted from Canaries) that are run on all PRs

This resolves an issue where URLs were being interpreted incorrectly. Most notably, WP core library urls (e.g. to files like `/wp-includes/js/dist/interactivity.min.js`) were being rendered as `https://dev-cxr-wpcm-canary-test.pantheonsite.io/wp-includes/js/dist/interactivity.min.js/?ver=6.5.5`.

Additionally, this should resolve the issues with URL structure causing the canary tests to fail.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -62,7 +62,7 @@ jobs:
           git config --global user.name "Pantheon WPCM Bot 🤖"
           PR_NUMBER=$(echo ${{ github.event.pull_request.number }})
           echo "Pull Request Number: ${PR_NUMBER}"
-          COMMIT_MSG=$(gh pr view ${PR_NUMBER} --json commits --jq '.commits[-1].messageHeadline')
+          COMMIT_MSG=$(gh pr view ${PR_NUMBER} --json commits --jq '.commits[-1] | "\(.messageHeadline) \(.messageBody)"')
           echo "Commit Message: ${COMMIT_MSG}"
           terminus local:clone wpcm-playwright-tests
           cd ~/pantheon-local-copies/wpcm-playwright-tests
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -77,7 +77,7 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           git config --global user.email "cms-platform+sage-testing@pantheon.io"
-          git config --global user.name "Pantheon WPCM Bot 🤖"
+          git config --global user.name "Pantheon WPCM Bot"
           PR_NUMBER=$(echo ${{ github.event.pull_request.number }})
           echo "Pull Request Number: ${PR_NUMBER}"
           COMMIT_MSG=$(gh pr view ${PR_NUMBER} --json commits --jq '.commits[-1] | "\(.messageHeadline) \(.messageBody)"')
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -55,6 +55,8 @@ jobs:
           fi
 
       - name: Clone the site and copy PR updates
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           git config --global user.email "cms-platform+sage-testing@pantheon.io"
           git config --global user.name "Pantheon WPCM Bot 🤖"
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -17,6 +17,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set up cache for dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.composer/cache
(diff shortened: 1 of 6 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Create Site
+        run: |
+          if terminus site:info wpcm-playwright-tests; then
+            echo "Test site already exists, skipping site creation."
+            # If the site exists already, we should switch it to git mode.
+            terminus connection:set wpcm-playwright-tests.dev git -y
+          else
+            terminus site:create wpcm-playwright-tests 'WordPress (Composer Managed) Playwright Test Site' 'WordPress (Composer Managed)' --org=5ae1fa30-8cc4-4894-8ca9-d50628dcba17
+          fi
```
pwtyler, line 73: site name and org UUID can/should remain config'd in this file, but can they be abstracted to global variables to ease identification (and replacement)?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Clone the site locally and copy PR updates
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "cms-platform+sage-testing@pantheon.io"
+          git config --global user.name "Pantheon WPCM Bot 🤖"
```
pwtyler, line 80:
```suggestion
          git config --global user.name "Pantheon WPCM Bot"
```
emojis cause surprises

web/app/mu-plugins/filters.php
```diff
@@ -96,6 +96,25 @@ function fix_core_resource_urls( string $url ) : string {
...
+	$options = get_option( 'graphql_general_settings' );
+	$site_path = site_url();
+
+	if ( ! $options ) {
+		$endpoint = ( ! empty( $site_path ) || strpos( $site_path, 'wp' ) !== false ) ? 'graphql' : 'wp/graphql';
+		$options = [];
+		$options['graphql_endpoint'] = $endpoint;
+		update_option( 'graphql_general_settings', $options );
+	}
+}
```
pwtyler, line 115:
```suggestion
function prepopulate_graphql_endpoint_url() {
	$options = get_option( 'graphql_general_settings' );
	if ($options) {
		return
	}
	$site_path = site_url();
	$endpoint = ( ! empty( $site_path ) || strpos( $site_path, 'wp' ) !== false ) ? 'graphql' : 'wp/graphql';
	$options = [
		'graphql_endpoint' => $endpoint
	];
	update_option( 'graphql_general_settings', $options );
}
```

web/app/mu-plugins/filters.php
```diff
@@ -172,15 +200,31 @@ function __is_login_url( string $url ) : bool {
...
-    // Reconstruct and return the full normalized URL.
-    return $scheme_with_separator . $normalized_url;
+/**
+ * Rebuild parsed URL from parts.
+ *
+ * @since 1.1.0
+ * @param array $parts URL parts from parse_url.
+ * @return string Re-parsed URL.
+ */
+function __rebuild_url_from_parts( array $parts ) : string {
```
pwtyler, line 222: feels very dirty that this doesn't exist already 😬

web/app/mu-plugins/filters.php
```diff
@@ -172,15 +200,31 @@ function __is_login_url( string $url ) : bool {
...
+	// Normalize the URL to remove any double slashes.
+	if ( isset( $parts['path'] ) ) {
+		$parts['path'] = preg_replace( '#/+#', '/', $parts['path'] );
+	}
 
-    // Normalize the remaining URL to remove any double slashes.
-    $normalized_url = str_replace( '//', '/', $remaining_url );
+	// Rebuild and return the full normalized URL.
+	return __rebuild_url_from_parts( $parts );
+}
```
pwtyler, line 213: seems over-engineered with the same result but maybe I'm missing what's changed and how the bugs manifested.

pwtyler (approved):
None of my feedback is a hard-block.

web/app/mu-plugins/filters.php
jazzsequence: I'm not entirely convinced it's necessary but I'm also not convinced that it's not. It was one of the many steps on the path of resolving the bug(s).

.github/workflows/playwright.yml
jazzsequence: Yes, I was just avoiding it to avoid complexity. We can come back to this to refine it later, I was more concerned that we _had_ tests.

PR #139, opened 2024-07-10
[BUGS-8491] only set `DISABLE_WP_CRON` value once

jazzsequence:
This PR solves several issues.

First and foremost, it resolves an issue created by the fact that `DISABLE_WP_CRON` was defined _twice_ (once in `application.php` and once in `wp-config-pantheon.php`).

The line in `application.php` has been removed.

The second important change here is that we've moved everything from `wp-config-pantheon.php` into `config/application.pantheon.php`. This puts all the Pantheon-specific settings in one place, in the `config` folder (where Bedrock likes it). The `wp-config-pantheon.php` file is retained for legacy reasons but the require is removed from `wp-config.php` (which should not be edited by users on Bedrock sites). The `DISABLE_WP_CRON` constant is defined in `application.pantheon.php` as a value that can be overridden if set before the file is loaded in `application.php` (which allows users the ability to change it if they really need to). Inline comments have been added to reiterate that this is _not_ recommended as well as points out that we do _not_ disable WP Cron on multisite (which the previous comment did not reference at all).

### DIFF
--- a/web/wp-config-pantheon.php
+++ b/web/wp-config-pantheon.php
@@ -1,70 +1,9 @@
 <?php
 /**
- * Pantheon platform settings.
+ * Pantheon wp-config file.
  *
- * IMPORTANT NOTE:
- * Do not modify this file. This file is maintained by Pantheon.
+ * This file is an artifact and has no content. It is not loaded by wp-config (Pantheon-specific configurations have been moved to config/application.pantheon.php).
  *
- * Site-specific modifications belong in wp-config.php, not this file.
- * This file may change in future releases and modifications would cause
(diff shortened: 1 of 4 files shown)


web/wp-config-pantheon.php
pwtyler, line 1: as customers should not be editing this file, I think this is safe to delete in the same PR that removes it from includes. Also this is EA, breaking changes are expected.

jazzsequence: 👍 ✅

pwtyler (approved):

PR #142, opened 2024-07-18
[CMSP-1140] fix subdirectory paths

jazzsequence:
This pull request removes the unnecessary code for handling wp-admin URLs. The code was not working as intended and caused issues.

This has been tested on a subdirectory multisite (which was the only thing that was broken or affected). Subdomain multisite should be unaffected (there are conditionals specifically targeting subdirectory multisite) but will be tested as well assuming automated tests pass.

The primary issue was admin URLs which are not (currently) tested with automated testing.

blocked by #140


jazzsequence:
Ohhhh, the robot should not change composer files....

jazzsequence:
not sure why this is pulling the full history. closing and reopening a clean pr

PR #146, opened 2024-07-30
[SITE-1585] Move check-commits into GitHub actions

jazzsequence:
This PR moves the `check-commits` script into a GitHub action workflow. It does not remove it from the deploy workflow, but adding it to a GHA workflow means that if we would mix non-release and release commits in a single PR, we are alerted early because the PR will fail.

This PR also creates a relationship between the Playwright (functional) tests and the linting tests and commit checks. If anything in the linting tests (`ci.yml`) fails, it prevents the Playwright tests from executing.

If the check-commits script fails -- and finds mixed commits to release and non-release files in the same PR -- a message is added to the PR and the workflow fails.

This does not change or fix the issues in the upstream or solve the issue described in SITE-1585, but it will (attempt to) prevent issues that currently exist in the upstream from happening again by failing the PR early.

### DIFF
--- a/devops/scripts/commit-type.sh
+++ b/devops/scripts/commit-type.sh
@@ -35,7 +35,7 @@ function identify_commit_type() {
 # Verifies that the given commit does not contain forbidden files.
 function only_allowed_files() {
   local commit=$1
-  local forbidden_files=("composer.lock" ".github/workflows/ci.yml")
+  local forbidden_files=("composer.lock" "package-lock.json")
   local has_forbidden_files=0
 
   affected_paths=$(git show "${commit}" --pretty=oneline --name-only | tail -n +2)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,8 @@ jobs:
 
     - name: Check Commits
       run: |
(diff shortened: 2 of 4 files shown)


devops/scripts/commit-type.sh
```diff
@@ -35,7 +35,7 @@ function identify_commit_type() {
 # Verifies that the given commit does not contain forbidden files.
 function only_allowed_files() {
   local commit=$1
-  local forbidden_files=("composer.lock" ".github/workflows/ci.yml")
+  local forbidden_files=("composer.lock" "package-lock.json")
```
greg-1-anderson, line 38: Why remove .github/workflows/ci.yml?

greg-1-anderson (approved):
LGTM.

devops/scripts/commit-type.sh
jazzsequence: For the immediate need, it was blocking the workflow from running on _this_ PR. However, I tried to figure out why we added it in the first place and it looked like a "¯\_(ツ)_/¯ maybe this will fix something" addition. There was no ticket associated with it and I don't really know why we would be disallowing changes to the workflow (they're non-release changes anyway).

PR #147, opened 2024-07-31
[SITE-1585] address diffs after rebase

jazzsequence:
This PR re-applies fixes from the backed-up copy of WPCM onto files that were changed the wrong way after the rebase.

While this PR doesn't actually address anything related to SITE-1585, rebasing the entire repository on the `pantheon-upstreams` repo, the "fix" was rebasing the repository on `pantheon-usptreams/main` and this fixes files that were adversely affected by the rebase (honestly surprising more didn't break).

PR #148, opened 2024-07-31
Get subdomain multisite tests working properly

jazzsequence:
We're not running the full setup-playwright-tests script because subdomain multisites don't need most of the setup. However, it does need _some_ of the setup (specifically copying the changes from the PR to the fixture site), so we need to manually add those back into that workflow.

In the future we might be able to run just portions of the script for the purpose of subdomain multisite so we're not duplicating code here.

~This PR is blocked by #147 -- the changes the robot fixed were part of what was updated in that PR. Merging this PR before that one will create a commit that gets dropped during deploy.~

This change includes some tweaks to the actual playwright tests and some deviations in the setup. It was discovered in running the tests that the subdomain fixture site was getting reset (GraphQL plugin being deleted, multisite config resetting) -- those have been added to the setup -- but in troubleshooting those it was hypothesized that an incorrect GraphQL endpoint url _could have been_ the issue (it wasn't, multisite not being set up was the issue), or that the environment variables weren't being processed correctly (that wasn't it either, but that was only discovered after breaking the Playwright tests into separate tests -- one for the main site and one for the subdomain). An additional passed environment variable is now supported, however, for GraphQL endpoint, which means that we can eventually remove the handling in the test for different endpoint urls for _subdirectory_ multisite (not included in this PR as it doesn't relate to the subdomain testing).

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -439,9 +439,7 @@ jobs:
             exit 1
           fi
           # Check SUBDOMAIN_INSTALL value
-          SUBDOMAIN_INSTALL=$(terminus wp "${{ env.SITE_ID }}".dev -- config get SUBDOMAIN_INSTALL)
-          if [[ "${SUBDOMAIN_INSTALL}" != "1" ]]; then
-            # SUBDOMAIN_INSTALL should be true.
+          if ! terminus wp "${{ env.SITE_ID }}".dev -- config is-true SUBDOMAIN_INSTALL; then
               echo -e "Subdomain configuration not found!"
               exit 1
           fi
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -400,7 +400,7 @@ jobs:
           terminus local:clone ${{ env.SITE_ID }}
           cd ~/pantheon-local-copies/"${{ env.SITE_ID }}"
           echo "Copying latest changes and committing to the site."
-          rsync -a --exclude='.git' --exclude='status-*.txt' "${{ github.workspace }}/" .
+          rsync -a --exclude='.git' --exclude='status-*.txt' --exclude="node_modules" "${{ github.workspace }}/" .
           git add -A
           git commit -m "Update to latest commit: ${{ env.COMMIT_MSG }}" || true
           git push origin master || true
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -393,6 +393,9 @@ jobs:
           git config --global user.email "cms-platform+sage-testing@pantheon.io"
           git config --global user.name "Pantheon WPCM Bot"
 
+          echo "Switching to git mode..."
+          terminus connection:set "${{ env.SITE_ID }}".dev git
+
           echo "Clone the site locally and copy PR updates"
           terminus local:clone ${{ env.SITE_ID }}
           cd ~/pantheon-local-copies/"${{ env.SITE_ID }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/playwright.yml
```diff
@@ -385,6 +386,37 @@ jobs:
...
+            echo -e "${RED}Multisite not found!${RESET}"
+            exit 1
+          fi
+          # Check SUBDOMAIN_INSTALL value
+          SUBDOMAIN_INSTALL=$(terminus wp "${{ env.SITE_ID }}".dev -- config get SUBDOMAIN_INSTALL)
+          if [[ "${SUBDOMAIN_INSTALL}" != "1" ]]; then
+            # SUBDOMAIN_INSTALL should be true.
+              echo -e "${RED}Subdomain configuration not found!${RESET}"
+              exit 1
+          fi
```
pwtyler, line 417:
```suggestion
          if ! terminus wp "${{ env.SITE_ID }}".dev -- config is-true SUBDOMAIN_INSTALL; then
              echo -e "Subdomain configuration not found!"
              exit 1
          fi
```

.github/workflows/playwright.yml
```diff
@@ -385,6 +386,37 @@ jobs:
...
+          git add -A
+          git commit -m "Update to latest commit: ${{ env.COMMIT_MSG }}" || true
+          git push origin master || true
+
+          echo "Checking WordPress install status"
+          terminus wp "${{ env.SITE_ID }}".dev -- cli info
+          if ! terminus wp "${{ env.SITE_ID }}".dev -- config is-true MULTISITE; then
+            echo -e "${RED}Multisite not found!${RESET}"
+            exit 1
+          fi
```
pwtyler, line 410:
```suggestion
          if ! terminus wp "${{ env.SITE_ID }}".dev -- config is-true MULTISITE; then
            echo -e "Multisite not found!"
            exit 1
          fi
```

pwtyler (approved):

jazzsequence:
@pwtyler re-requesting review since I've made changes since your ✅

.github/workflows/playwright.yml
```diff
@@ -385,13 +401,63 @@ jobs:
...
+          terminus env:clear-cache "${{ env.SITE_ID }}".dev
+
+          # Run curl checks against the site URLs to ensure the main site and the subdomain site exist.
+          echo "Checking site URLs"
+          SITE_URL_TEST=$(curl -s -o /dev/null -w "%{http_code}" ${{ env.SITE_URL }})
+          SUBDOMAIN_URL_TEST=$(curl -s -o /dev/null -w "%{http_code}" ${{ env.SUBDOMAIN_URL }})
+          if [ $SITE_URL_TEST -ne 200 ] || [ $SUBDOMAIN_URL_TEST -ne 200 ]; then
+            echo "One or more site URLs are not returning a 200 status code. Exiting."
+            exit 1
+          fi
```
pwtyler, line 451: might be useful to debug print what each are
```suggestion
          if [ $SITE_URL_TEST -ne 200 ] || [ $SUBDOMAIN_URL_TEST -ne 200 ]; then
            echo "One or more site URLs are not returning a 200 status code. Exiting."
            echo "${{ env.SITE_URL }} - ${SITE_URL_TEST}"
            echo "${{ env.SUBDOMAIN_URL }} - ${SUBDOMAIN_URL_TEST}"
            exit 1
          fi
```

pwtyler (approved):

PR #152, opened 2024-08-02
Handle merge commits with multiple parents and conflicts

jazzsequence:
This pull request adds handling for merge commits with multiple parents and conflicts where the cherry-picked commit delets a file that does not exist in the `--theirs` version.

Test deploy is here: https://github.com/pantheon-upstreams/wordpress-composer-managed/compare/main...jazzsequence:wordpress-composer-managed:main

https://github.com/pantheon-systems/wordpress-composer-managed/pull/152/commits/78e48620916e213fe0a4c57698de764871d49130 was added because it looks like the `composer.lock` was being added in that deploy https://github.com/jazzsequence/wordpress-composer-managed

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -9,6 +9,7 @@ on:
       - '.github/workflows/sync-default.yml'
       - '.github/tests/*.bats'
       - 'private/scripts/**'
+      - 'devops/**'
       - 'docs/**'
       - '*.md'
       - 'phpcs.yml'
--- a/devops/scripts/deploy-public-upstream.sh
+++ b/devops/scripts/deploy-public-upstream.sh
@@ -79,9 +79,7 @@ for commit in "${commits[@]}"; do
 
     # Stage the changes and continue the cherry-pick
     git add -u
-    git commit --no-edit || {
-      echo "No changes to commit. Continuing."
-    }
+    git commit --no-edit || echo "No changes to commit. Continuing."
   fi
   # Product request - single commit per release
   # The commit message from the last commit will be used.
--- a/devops/scripts/deploy-public-upstream.sh
+++ b/devops/scripts/deploy-public-upstream.sh
@@ -68,7 +68,26 @@ for commit in "${commits[@]}"; do
   fi
   echo "Adding $commit:"
   git --no-pager log --format=%B -n 1 "$commit"
-  git cherry-pick -rn -X theirs -m 1 "$commit" 2>&1
+  if ! git cherry-pick -rn -X theirs -m 1 "$commit"; then
+    echo "Conflict detected in $commit. Checking for deleted files."
+    conflicted_files=$(git diff --name-only --diff-filter=U)
+    for file in $conflicted_files; do
+      if git show "$commit" -- "$file" | grep -q 'delete mode'; then
+        echo "File $file was deleted in the cherry-picked commit. Resolving by keeping the deletion."
+        git rm "$file"
+      else
+        echo "Resolving conflict for $file with 'theirs'."
+        git checkout --theirs -- "$file"
+        git add "$file"
+      fi
+    done
+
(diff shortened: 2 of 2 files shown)


.github/workflows/playwright.yml
```diff
@@ -9,6 +9,7 @@ on:
       - '.github/workflows/sync-default.yml'
       - '.github/tests/*.bats'
       - 'private/scripts/**'
+      - 'devops/**'
```
jazzsequence, line 12: Making a change to this file explicitly does _not_ exclude this PR from running these tests because we just made a change to `playwright.yml` which triggers the workflow. `¯\_(ツ)_/¯`

devops/scripts/deploy-public-upstream.sh
```diff
@@ -65,7 +65,24 @@ for commit in "${commits[@]}"; do
...
-  git cherry-pick -rn -X theirs "$commit" 2>&1
+  if ! git cherry-pick -rn -X theirs -m 1 "$commit"; then
+    echo "Conflict detected in $commit. Checking for deleted files."
+    conflicted_files=$(git diff --name-only --diff-filter=U)
+    for file in $conflicted_files; do
+      if ! git ls-tree -r "$commit" --name-only | grep -q "^$file$"; then
+        echo "File $file was deleted in the cherry-picked commit. Resolving by keeping the deletion."
+        git rm "$file"
+      else
+        echo "Conflict required manual resolution for $file."
```
pwtyler, line 76: does this line need to bail (or set a flag to bail after)?

devops/scripts/deploy-public-upstream.sh
```diff
@@ -65,7 +65,24 @@ for commit in "${commits[@]}"; do
...
+      fi
+    done
+
+    # Stage the changes and continue the cherry-pick
+    git add -u
+    git commit --no-edit || {
+      echo "No changes to commit. Continuing."
+    }
```
pwtyler, line 84:
```suggestion
    git commit --no-edit || echo "No changes to commit. Continuing."
```
not familiar with what the brackets are doing here was that supposed to be a subshell `()`?

devops/scripts/deploy-public-upstream.sh
```diff
@@ -75,6 +92,18 @@ echo "Copying README to docroot."
 rm ./README.md
 cp /tmp/composer-managed-README.md ./README.md
 
+# Create a list of files that we need to exclude from the commit.
+# These are files that may be coming from Roots or old commits that do not exist in the source and _should not exist_ on the upstream.
+ignored_files="composer.lock CODE_OF_CONDUCT.md CONTRIBUTING.md"
+
+# Remove ignored files from the commit.
+for file in $ignored_files; do
```
pwtyler, line 100: this should be an array not a space delineated string but I care 2/10

devops/scripts/deploy-public-upstream.sh
jazzsequence: It should bail on its own. When I was running it locally it caused the script to halt, which in Circle should cause the workflow to fail. I don't think we need a more explicit bail here.

devops/scripts/deploy-public-upstream.sh
jazzsequence: ¯\_(ツ)_/¯ it was a chatgpt suggestion. yours is easier to grok

devops/scripts/deploy-public-upstream.sh
jazzsequence: I thought space separated string _was_ how you did arrays in bash. Historically, we ran into issues creating arrays (which probably means we were __doing_it_wrong) and this is how I remember doing it so it worked.

pwtyler: this works, its just not ideal

pwtyler (approved):

PR #167, opened 2024-09-23
Rename catalog-info.yml to catalog-info.yaml

namespacebrian:
Automatically renamed catalog-info.yml to catalog-info.yaml


greg-1-anderson:
We should make sure this file isn't published to pantheon-upstreams before merging. Probably requires a small update to the block list in the release automation scripts.

PR #173, opened 2025-01-31
[SITE-3578] Fix missing index php

jazzsequence:
In some cases, we have gotten reports that the `web/index.php` goes missing. When this happens, it breaks the site because we also ignore `web/index.php` in the `.gitignore` file.

I have not been able to reproduce the circumstances that make this happen, however, if I manually delete `web/index.php`, it does, indeed, break the site irreparably (since we point to `/wp/wp-blog-header.php` instead of just `/wp-blog-header.php`).

We could solve the issue by removing the line in the `.gitignore` file but as it doesn't seem to be having this effect normally, and because that would potentially break anyone who has customized their `.gitignore`, this is not desirable.

Instead, this PR adds a check for `web/index.php` when we're checking if we need to, and creating symlinks to files in the `/wp` directory and are already in the `/web` directory. If `index.php` does not already exist, the additional function creates the file.

A BATS tests has been added to validate the behavior.

### DIFF
--- a/.github/tests/1-test-update-php.bats
+++ b/.github/tests/1-test-update-php.bats
@@ -76,3 +76,11 @@ teardown() {
   run grep -q "php_version: 8.3" pantheon.yml
   [ "$status" -eq 0 ]
 }
+
+@test "Test missing index.php" {
+  # Remove web/index.php.
+  rm web/index.php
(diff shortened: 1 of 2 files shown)


jazzsequence:
Failing tests are related to Sage Install. I will investigate these separately but they are unrelated to this PR.

pwtyler (approved):

PR #175, opened 2025-02-10
[SITE-3292] add @maybe-add-symlinks to post-update-cmd

jazzsequence:
Add the symlinks if they don't already exist.

When applying the upstream updates from 1.32.4, it was discovered that the symlinks to `wp-*` files (and the `index.php` file creation) is not run leaving the site in an empty state. When cache expires, the site will be broken with the files missing. (Perhaps this was the issue all along?)

This uses the existing `ComposerScripts` class and `preUpdate` hook (hooked to the composer `pre-update-cmd`) to update the `composer.json` to add `@maybe-install-symlinks` if that's not already part of the `post-update-cmd`. The `post-update-cmd` is also created by this script (if it did not exist already), so this change checks for the existance of the script we're adding and adds it if it's not there, which is then run after the `composer update` is executed.


pwtyler (approved):

jnewell-docker:
@jazzsequence Should it be `@maybe-create-symlinks` instead of `@maybe-install-symlinks`?

PR #177, opened 2025-02-10
set service level to pro on site creation for playwright tests

### LINKED ISSUE #176: Multisite Playwright tests fail if site is frozen
If a site is frozen, the playwright tests will fail because there's no site available that can be tested. (see https://github.com/pantheon-systems/wordpress-composer-managed/actions/runs/13248341593/job/36980142952)

Whether a site is frozen or not can be determined with `terminus site:info` (is_frozen: true|false), and a site can be "thawed" with `terminus site:thaw`.

~The workflow should be updated to check if the site is frozen and thaw it if it is.~

Alternative solution: `terminus service-level:set` can be used to set the site to a `pro` level on `site:create` to ensure it does not get frozen in the future.

In the meantime, existing fixture sites can be set to a non-sandbox site plan.

jazzsequence:
fixes #176

Sets the site plan to `pro` on site creation.

This applies to subdir and single site fixture sites. the subdomain fixture site is a standing site (which has already had its service level set to `pro`).

No release necessary for this change as it is CI only.

### DIFF
--- a/devops/scripts/setup-playwright-tests.sh
+++ b/devops/scripts/setup-playwright-tests.sh
@@ -39,6 +39,8 @@ create_site() {
     echo "Test site already exists, skipping site creation."
   else
     terminus site:create "${site_id}" "${site_name}" "${UPSTREAM_NAME}" --org=5ae1fa30-8cc4-4894-8ca9-d50628dcba17
+    echo "Site created. Setting site plan to 'pro'"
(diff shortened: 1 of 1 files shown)


jazzsequence:
Can and should hold this until after 1.32.5.

pwtyler (approved):

PR #181, opened 2025-04-22
[DEVREL-29] allow pretty permalinks to be handled for rest endpoints by default

### LINKED ISSUE #178: 404 errors when accessing wp-json from WPMS subsites
See https://github.com/pantheon-systems/documentation/issues/9500 and linked internal document comment.

> If creating a multisite from WordPress Composer Managed upstream, users may encounter 404 errors when saying pages or doing API related functions.
>
> This is due to the composer path /wp being wrongly prepended on endpoints. Example:
>
> When you try to save a post/page, inspecting the page would show that it is requests on /wp-json/wp/v2/settings?_locale=user. Instead, it should be /wp/wp-json/v2/settings?_locale=user
>
> From @ miriamgoldman: 
>>
>>Solution would be to filter the /wp-json paths to match having /wp on the right spot.
>>
>> ```add_filter('rest_url', function ($url) {
>>    return str_replace(home_url('/wp-json'), home_url('/wp/wp-json'), $url);
>> });
>> ```

This is a _bug_ in `filters.php` and should be fixed in the upstream rather than documenting or recommending a workaround.

jazzsequence:
Fixes #178 

The linked issue is somewhat misrepresented. While the observed behavior is accurate (malformed URLs) it is _not_ because of a misplaced `/wp` in the REST url -- WP-API URLs on Bedrock-based sites are _expected_ to have two `/wp/` strings in them -- one being the path to the application and the second is part of the actual `wp-json/` URL (e.g. `/wp-json/wp/v2/...`). Therefore, the proposed solution would actually _break_ sites without addressing the issue.

Since the issue only exists on _newly created sites_, the assumption here is that these sites are _not using "pretty permalinks"_. Indeed, our tests _assume_ permalinks are pretty for this very reason. In cases when "pretty permalinks" are not active, it makes sense that API requests might fail due to our URL rewriting for internal paths.

This PR works around the issue by simply _accepting_ "pretty"-style API endpoint paths while also preserving the functionality of "plain" permalinks. Both "plain" and "pretty" permalinks are henceforth supported by the REST API for WPCM sites.

~Tests are broken out into #182 so as to not mix [release and non-release commits](https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/README-internal.md#release-and-non-release-commits) and will fail until this PR is merged.~

Tests have been integrated back into this PR as well as other non-release changes. As a result, this PR MUST BE merged with a merge commit, **not** squash merge.


pwtyler:
> Tests are broken out into https://github.com/pantheon-systems/wordpress-composer-managed/pull/182 so as to not mix [release and non-release commits](https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/README-internal.md#release-and-non-release-commits) and will fail until this PR is merged.

This is why "rebase and merge" exists. I'm going to cherry-pick them back over here.

web/app/mu-plugins/filters.php
```diff
@@ -235,3 +235,149 @@ function __rebuild_url_from_parts( array $parts ) : string {
...
+		if ( strpos( $extracted_route, 'wp/' ) === 0 ) {
+			$extracted_route = substr( $extracted_route, strlen( 'wp' ) ); // Remove the extra 'wp'
+		}
+		// Ensure the extracted route starts with a slash
+		if ( ! $extracted_route && $extracted_route[0] !== '/' ) {
+			$extracted_route = '/' . $extracted_route;
+		}
+		$rest_route = $extracted_route ?: '/'; // Use extracted route or default to base
+	}
+	return $rest_route;
```
pwtyler, line 265:
```suggestion
	if ( $wp_json_pos === false ) {
		return $rest_route;
	}
	$extracted_route = substr( $path, $wp_json_pos + strlen( '/wp-json' ) ); // Get everything after /wp-json
	// Special case: Handle the originally reported '/wp-json/wp/' malformation
	if ( strpos( $extracted_route, 'wp/' ) === 0 ) {
		$extracted_route = substr( $extracted_route, strlen( 'wp' ) ); // Remove the extra 'wp'
	}
	// Ensure the extracted route starts with a slash
	if ( ! $extracted_route && $extracted_route[0] !== '/' ) {
		$extracted_route = '/' . $extracted_route;
	}
	$rest_route = $extracted_route ?: '/'; // Use extracted route or default to base
```
disappointed. I thought you were on the return early train.

web/app/mu-plugins/filters.php
```diff
@@ -235,3 +235,149 @@ function __rebuild_url_from_parts( array $parts ) : string {
...
+		$endpoint = ($endpoint === '/') ? '' : $endpoint;
+
+		// Check if rest_route is already set (e.g., from query string), if so, don't overwrite.
+		// This prevents conflicts if someone manually crafts a URL like /wp/wp-json/posts?rest_route=/users
+		if ( ! isset( $wp->query_vars['rest_route'] ) ) {
+			// Directly set the query variable for the REST API.
+			$wp->query_vars['rest_route'] = $endpoint;
+
+			// Optional: Unset other query vars WP might have incorrectly parsed.
+			// unset($wp->query_vars['pagename'], $wp->query_vars['error']);
```
pwtyler, line 372: keeping this around? or delete
```suggestion
```

jazzsequence:
Going to close this PR and open a new one with two clean commits.

PR #182, opened 2025-04-22
[DEVREL-29] Adds testing for new URL handling for API endpoints when "pretty permalinks" aren't set

jazzsequence:
- [ ] Hold for #181 to be merged to `default`

This PR adds tests that validate that API endpoints can still be hit appropriately when a site is using "plain permalinks".


pwtyler:
Cherry-picked back over to #181, just need to rebase and merge instead of squashing that one.

PR #185, opened 2025-05-07
Release 1.33.0

jazzsequence:
> Fixes a Composer script error and an issue where WP REST API urls would be broken on newly created sites.
>
> For more information see https://docs.pantheon.io/release-notes/2025/05/wordpress-composer-managed-1-33-0


jazzsequence:
tests are currently failing because the `service-level` terminus command has been deprecated. This is fixed in #186 so when that's merged to `default`, the tests should padd.

PR #188, opened 2025-06-26
Feedback wanted: Integrate FAIR into WordPress (Composer Managed)

jazzsequence:
This PR follows on from https://github.com/pantheon-systems/WordPress/pull/419. Discussion about whether and why to integrate FAIR into WordPress upstreams broadly should use that ticket for discussion. The implementation is largely the same as that PR.

## Considerations

- **Integration into [Pantheon Update Tool](https://github.com/pantheon-systems/update-tool) is unnecessary/irrelevant.** That workflow exists to manage dependencies for non-Composer-based environments. With Composer, we can simply use Composer itself.
- **The FAIR plugin is not currently on Packagist.** This means that we need to add the source repository to the repositories list tin the `composer.json` file. There's a (strong) argument here to wait until it's added to Packagist.org before this PR is merged.
- **The FAIR plugin installs as a WordPress plugin by default.** This requires a user to activate the plugin from the plugins page in the WordPress admin. Our proposed approach is to install as an mu-plugin. To do this, we must change the `composer.json` file, but doing so (at least in the current implementation) can cause merge conflicts and/or rewrite the site `composer.json` file on existing WPCM sites. The safest way of applying this update is doing so [manually](https://docs.pantheon.io/guides/custom-upstream/troubleshooting#manually-resolve-from-the-command-line) (Note: current docs do not show how to do this with WPCM), which creates a barrier to entry.
- **The same issue exists with default state for avatars.** This could be addressed in an update to the Pantheon MU Plugin and pulled in automatically. See https://github.com/pantheon-systems/pantheon-mu-plugin/pull/91

Route feedback and discussion to https://github.com/pantheon-systems/WordPress/pull/419 and link to this issue to show that you are referring to the Composer Managed implementation of WordPress.


jazzsequence:
One thought I have about this that is possibly a counter-argument to inclusion is the fact that FAIR is largely a stand-in replacement for WordPress automatic/internal updates and install mechanisms. With Composer-based sites, we're getting updates from Composer (and, likely WPackagist in front of WordPress.org). Unless/until there is a stand-in integration/replacement within FAIR that steps in front of api.wordpress.org after wpackagist.org, we'll only get the limited value of the extra API stuff (internal solutions for A8c-based APIs/properties) that the FAIR plugin adds, but won't benefit from the protocol at all, really.

jazzsequence:
See https://github.com/orgs/fairpm/discussions/41 related to the above comment ☝️

PR #189, opened 2025-08-07
Resolve PHP 8.1 warning for rtrim() in functions.php

mattmacneil:
Got rid of rtrim() deprecation that was logged for each page load in my local dev environment. 

The issue was that trailingslashit() was sometimes getting a null value from parse_url(), which throwa a warning in php 8.1.

Updated line uses ?? operator so it always passes a string instead of null.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### v1.34.0 ()
+* Resolves rtrim warning in PHP 8.1+ environments. ([#189](https://github.com/pantheon-systems/wordpress-composer-managed/pull/189)) props @mattmacneil
+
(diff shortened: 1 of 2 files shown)


mattmacneil:
Hmm I opened this PR thinking this was an issue "out of the box" with the Composer managed WP template; now I'm thinking this is might be something specific to my theme (which I assumed was the case at first but later dismissed in error). 

Leaving this open in case it's still useful, but my assumption that everyone else using php 8.3 would constantly see this warning too is probably not correct.

thetwopct:
This worked for me 👍

jazzsequence (approved):

jazzsequence (approved):

PR #196, opened 2026-05-04
Sync with upstream Bedrock

retlehs:
This PR syncs upstream Bedrock changes into the Pantheon fork

- WPackagist → WP Packages
- Removed `roots/wp-password-bcrypt` (WordPress 6.8 ships native bcrypt password hashing)
- `config/application.php` changes:
    - **`Env\Env::$options` as named constants** — replaces the magic value `31` with explicit flags ([Bedrock reference](https://github.com/roots/bedrock/blob/0d38b039ed2f5937b52ca813ca9195de5f63b0cf/config/application.php#L16C1-L22)). Pantheon's flag set is preserved exactly: `USE_ENV_ARRAY` is kept (needed for the `$_ENV[...]` syntax in `.env.example`), and `LOCAL_FIRST` is intentionally not added (Pantheon's Lando-aware loading handles env precedence differently).
    - **`WP_ENVIRONMENT_TYPE`** — derives from env or falls back to `WP_ENV` if it matches a recognized value.
    - **`WP_DEVELOPMENT_MODE`** — supports the WP 6.3+ development-mode flag via env.
    - **`DB_SSL`** — opt-in MySQL SSL (`MYSQLI_CLIENT_SSL`), no-op unless the env var is set.
    - **`DISABLE_WP_CRON`** — env-configurable for sites running cron externally.
    - **`CONCATENATE_SCRIPTS = false`** — Bedrock convention to disable WP's built-in script concatenation.
- `composer.json` housekeeping
    - **Default theme**: `wpackagist-theme/twentytwentytwo:^1.2` → `wp-theme/twentytwentyfive:^1.0`. The previous theme is from WP 5.9 (Jan 2022); twentytwentyfive is the current default in WP 6.7+.
    - **`"sort-packages": true`** added under `config:` — keeps `composer.lock` diffs deterministic.
    - Both `require` and `require-dev` alphabetized so `sort-packages` is meaningful immediately, not just for future additions.

Happy to adjust anything.

**Disclosure:** I'm one of the [Bedrock](https://roots.io/bedrock/) maintainers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,8 +40,8 @@
   "repositories": [
     {
       "type": "composer",
-      "url": "https://wpackagist.org",
-      "only": ["wpackagist-plugin/*", "wpackagist-theme/*"]
+      "url": "https://repo.wp-packages.org",
+      "only": ["wp-plugin/*", "wp-theme/*"]
     },
     {
         "type": "path",
@@ -51,29 +51,29 @@
   "require": {
     "php": ">=8.0",
     "composer/installers": "^2.2",
-    "vlucas/phpdotenv": "^5.5",
+    "cweagans/composer-patches": "^1.7",
     "oscarotero/env": "^2.1",
+    "pantheon-systems/pantheon-mu-plugin": "*",
+    "pantheon-upstreams/upstream-configuration": "dev-main",
     "roots/bedrock-autoloader": "*",
     "roots/bedrock-disallow-indexing": "*",
     "roots/wordpress": "*",
     "roots/wp-config": "*",
-    "roots/wp-password-bcrypt": "*",
-    "wpackagist-theme/twentytwentytwo": "^1.2",
-    "pantheon-systems/pantheon-mu-plugin": "*",
-    "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
-    "cweagans/composer-patches": "^1.7"
+    "vlucas/phpdotenv": "^5.5",
+    "wp-plugin/pantheon-advanced-page-cache": "*",
+    "wp-plugin/wp-native-php-sessions": "*",
+    "wp-theme/twentytwentyfive": "^1.0"
   },
   "require-dev": {
+    "assertwell/shellcheck": "^1.0",
     "pantheon-systems/pantheon-wp-coding-standards": "*",
-    "roave/security-advisories": "dev-latest",
-    "assertwell/shellcheck": "^1.0"
+    "roave/security-advisories": "dev-latest"
   },
   "config": {
     "optimize-autoloader": true,
     "preferred-install": "dist",
     "process-timeout": 0,
+    "sort-packages": true,
     "allow-plugins": {
       "composer/installers": true,
       "roots/wordpress-core-installer": true,
--- a/config/application.php
+++ b/config/application.php
@@ -11,7 +11,12 @@ use Roots\WPConfig\Config;
 use function Env\env;
 
 // USE_ENV_ARRAY + CONVERT_* + STRIP_QUOTES.
-Env\Env::$options = 31;
+Env\Env::$options
+	= Env\Env::USE_ENV_ARRAY
+	| Env\Env::CONVERT_BOOL
+	| Env\Env::CONVERT_NULL
+	| Env\Env::CONVERT_INT
+	| Env\Env::STRIP_QUOTES;
 
 /**
  * Directory containing all of the site's files
@@ -60,8 +65,36 @@ require_once __DIR__ . '/application.pantheon.php';
 define( 'WP_ENV', env( 'WP_ENV' ) ?: 'production' );
 
 /**
+ * Set WP_ENVIRONMENT_TYPE if not already defined
+ */
+if ( ! defined( 'WP_ENVIRONMENT_TYPE' ) ) {
+	$wp_environment_type = env( 'WP_ENVIRONMENT_TYPE' );
+
+	if ( $wp_environment_type ) {
+		Config::define( 'WP_ENVIRONMENT_TYPE', $wp_environment_type );
+	} elseif ( in_array( WP_ENV, [ 'production', 'staging', 'development', 'local' ], true ) ) {
+		Config::define( 'WP_ENVIRONMENT_TYPE', WP_ENV );
+	}
+}
+
+/**
+ * Set WP_DEVELOPMENT_MODE if explicitly configured
+ */
+if ( ! defined( 'WP_DEVELOPMENT_MODE' ) ) {
+	$wp_development_mode = env( 'WP_DEVELOPMENT_MODE' );
+
+	if ( $wp_development_mode ) {
+		Config::define( 'WP_DEVELOPMENT_MODE', $wp_development_mode );
+	}
+}
+
+/**
  * DB settings
  */
+if ( env( 'DB_SSL' ) ) {
+	Config::define( 'MYSQL_CLIENT_FLAGS', MYSQLI_CLIENT_SSL );
+}
+
 Config::define( 'DB_NAME', env( 'DB_NAME' ) );
 Config::define( 'DB_USER', env( 'DB_USER' ) );
 Config::define( 'DB_PASSWORD', env( 'DB_PASSWORD' ) );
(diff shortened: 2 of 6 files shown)


config/application.php
```diff
@@ -116,12 +149,15 @@
  * Custom Settings
  */
 Config::define( 'AUTOMATIC_UPDATER_DISABLED', true );
+Config::define( 'DISABLE_WP_CRON', env( 'DISABLE_WP_CRON' ) ?: false );
```
jazzsequence: We explicitly leave this disabled (and this would be overridden by the mu-plugin) because we have a once hourly system cron. Users can use a custom cron job or turn this back on if they need it, but it's off by default.

config/application.php
```diff
@@ -116,12 +149,15 @@
...
 // Disable the plugin and theme file editor in the admin.
 Config::define( 'DISALLOW_FILE_EDIT', true );
 // Disable plugin and theme updates and installation from the admin.
 Config::define( 'DISALLOW_FILE_MODS', true );
 // Limit the number of post revisions that Wordpress stores (true (default WP): store every revision).
 Config::define( 'WP_POST_REVISIONS', env( 'WP_POST_REVISIONS' ) ?? true );
+// Disable script concatenation.
+Config::define( 'CONCATENATE_SCRIPTS', false );
```
jazzsequence: Not sure why this needs to be defined for everyone. This seems user-specific.

composer.json
```diff
@@ -51,29 +51,29 @@
...
-    "wpackagist-theme/twentytwentytwo": "^1.2",
-    "pantheon-systems/pantheon-mu-plugin": "*",
-    "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
-    "cweagans/composer-patches": "^1.7"
+    "vlucas/phpdotenv": "^5.5",
+    "wp-plugin/pantheon-advanced-page-cache": "*",
+    "wp-plugin/wp-native-php-sessions": "*",
+    "wp-theme/twentytwentyfive": "^1.0"
```
jazzsequence: This should be left at the old theme. Changing in the upstream means anyone who was using this theme will suddenly find it uninstalled and their site broken.

jazzsequence:
```suggestion
    "wp-theme/twentytwentytwo": "^1.2"
```

jazzsequence (commented):
This is almost certainly going to cause merge conflicts as soon as people pull it down. That said, it might not be avoidable if we're changing the repository source anyway. 

@pwtyler I'd like your thoughts. I'm broadly in favor of switching from wpackagist to wp-packages, and if we're modifying the composer.json anyway (which is where I expect the merge conflicts to be), adding sort-packages and alphabetizing seems better to do it now then later.

config/application.php
jazzsequence:
```suggestion
```

config/application.php
jazzsequence:
```suggestion
```

composer.json
```diff
@@ -51,29 +51,29 @@
   "require": {
     "php": ">=8.0",
     "composer/installers": "^2.2",
-    "vlucas/phpdotenv": "^5.5",
+    "cweagans/composer-patches": "^1.7",
```
jazzsequence: Reverting to the non-alphabetized version to reduce the number of merge conflicts this upstream update will cause.

```suggestion
    "vlucas/phpdotenv": "^5.5",
```

composer.json
```diff
@@ -51,29 +51,29 @@
...
     "oscarotero/env": "^2.1",
+    "pantheon-systems/pantheon-mu-plugin": "*",
+    "pantheon-upstreams/upstream-configuration": "dev-main",
```
jazzsequence:
```suggestion
```

composer.json
```diff
@@ -51,29 +51,29 @@
...
-    "wpackagist-theme/twentytwentytwo": "^1.2",
-    "pantheon-systems/pantheon-mu-plugin": "*",
-    "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
-    "cweagans/composer-patches": "^1.7"
+    "vlucas/phpdotenv": "^5.5",
+    "wp-plugin/pantheon-advanced-page-cache": "*",
+    "wp-plugin/wp-native-php-sessions": "*",
+    "wp-theme/twentytwentytwo": "^1.2"
```
jazzsequence:
```suggestion
    "wp-theme/twentytwentytwo": "^1.2"
    "pantheon-systems/pantheon-mu-plugin": "*",
    "pantheon-upstreams/upstream-configuration": "dev-main",
    "wp-plugin/pantheon-advanced-page-cache": "*",
    "wp-plugin/wp-native-php-sessions": "*",
```

composer.json
```diff
@@ -51,29 +51,29 @@
...
   },
   "require-dev": {
+    "assertwell/shellcheck": "^1.0",
```
jazzsequence:
```suggestion
```

composer.json
```diff
@@ -51,29 +51,29 @@
...
     "pantheon-systems/pantheon-wp-coding-standards": "*",
-    "roave/security-advisories": "dev-latest",
-    "assertwell/shellcheck": "^1.0"
+    "roave/security-advisories": "dev-latest"
```
jazzsequence:
```suggestion
    "roave/security-advisories": "dev-latest",
    "assertwell/shellcheck": "^1.0"
```

composer.json
```diff
@@ -51,29 +51,29 @@
...
   },
   "config": {
     "optimize-autoloader": true,
     "preferred-install": "dist",
     "process-timeout": 0,
+    "sort-packages": true,
```
jazzsequence:
```suggestion
```

composer.json
```diff
@@ -57,13 +57,11 @@
...
     "roots/wp-config": "*",
-    "roots/wp-password-bcrypt": "*",
-    "wpackagist-theme/twentytwentytwo": "^1.2",
+    "wp-theme/twentytwentytwo": "^1.2"
     "pantheon-systems/pantheon-mu-plugin": "*",
     "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
-    "cweagans/composer-patches": "^1.7"
+    "wp-plugin/pantheon-advanced-page-cache": "*",
```
jazzsequence:
```suggestion
    "cweagans/composer-patches": "^1.7"
    "wp-plugin/pantheon-advanced-page-cache": "*",
```

composer.json
```diff
@@ -57,13 +57,12 @@
...
-    "roots/wp-password-bcrypt": "*",
-    "wpackagist-theme/twentytwentytwo": "^1.2",
+    "wp-theme/twentytwentytwo": "^1.2"
     "pantheon-systems/pantheon-mu-plugin": "*",
     "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
     "cweagans/composer-patches": "^1.7"
+    "wp-plugin/pantheon-advanced-page-cache": "*",
+    "wp-plugin/wp-native-php-sessions": "*",
```
jazzsequence:
```suggestion
```

composer.json
```diff
@@ -57,13 +57,12 @@
...
     "roots/wordpress": "*",
     "roots/wp-config": "*",
-    "roots/wp-password-bcrypt": "*",
-    "wpackagist-theme/twentytwentytwo": "^1.2",
+    "wp-theme/twentytwentytwo": "^1.2"
     "pantheon-systems/pantheon-mu-plugin": "*",
     "pantheon-upstreams/upstream-configuration": "dev-main",
-    "wpackagist-plugin/pantheon-advanced-page-cache": "*",
-    "wpackagist-plugin/wp-native-php-sessions": "*",
     "cweagans/composer-patches": "^1.7"
```
jazzsequence:
```suggestion
    "wp-plugin/pantheon-advanced-page-cache": "*",
    "wp-plugin/wp-native-php-sessions": "*",
    "cweagans/composer-patches": "^1.7"
```

composer.json
```diff
@@ -57,12 +57,11 @@
     "roots/bedrock-disallow-indexing": "*",
     "roots/wordpress": "*",
     "roots/wp-config": "*",
-    "roots/wp-password-bcrypt": "*",
-    "wpackagist-theme/twentytwentytwo": "^1.2",
+    "wp-theme/twentytwentytwo": "^1.2"
```
jazzsequence:
```suggestion
    "wp-theme/twentytwentytwo": "^1.2",
```

jazzsequence:
Functional tests are failing for unrelated reasons. I'm also wondering if we can split the release and non-release commits into separate PRs so this is cleaner (see https://github.com/pantheon-systems/wordpress-composer-managed/pull/200#issuecomment-5780973548 which doesn't post here since this is a forked PR).

jazzsequence:
closing in favor of #201 which breaks the release/non-release changes into separate commits

